PR #10, opened 2020-08-10
Support dotnet

aaronpowell:
This PR supports dotnet as a SWA host environment, for doing blazor apps, but it also lays the groundwork to extend for other host runtimes going forward.

There is a bit of a rollback on the current functionality with regards to how the requests to the web host works. With a dotnet app, we rely on the Blazor dev server to handle some file pathing, which changes at publish (especially for auth). Also, since it's a compiled app, it's preferable to run the dotnet server, in watch mode, then just rely on the published files.

I've tested this with both blazor and JavaScript apps locally.

### DIFF
--- a/bin/index.js
+++ b/bin/index.js
@@ -107,7 +107,10 @@ if (program.ui) {
     });
 
   // start hosting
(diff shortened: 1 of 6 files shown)


aaronpowell:
Closing in favour of #11

Issue #6, opened 2020-07-30
Emulate routes config


manekinekko:
Checkout [docs](https://docs.microsoft.com/azure/static-web-apps/routes) for more details.

aaronpowell:
With #11 I've implemented a first-pass on supporting this.

When the proxy starts up it'll look for a `routes.json` file in the `app_artifact_location`, and if it exists it'll load those routing rules into the proxy server. Then, when a request comes in, the proxy server will check if the route matches a rule. If a rule is found, the app which check what the rule states (access control, redirect or status code response), and then apply what the rule states.

aaronpowell:
There's still a few outstanding things:

- [ ] [Custom error pages](https://docs.microsoft.com/en-gb/azure/static-web-apps/routes?WT.mc_id=-github-wachegha#custom-error-pages)
- [ ] [Custom mime types](https://docs.microsoft.com/en-gb/azure/static-web-apps/routes?WT.mc_id=-github-wachegha#custom-mime-types)
- [ ] [Default headers](https://docs.microsoft.com/en-gb/azure/static-web-apps/routes?WT.mc_id=-github-wachegha#default-headers)
- [ ] [Redirects](https://docs.microsoft.com/en-gb/azure/static-web-apps/routes?WT.mc_id=-github-wachegha#redirects) (it won't send the right status code response)

anthonychu:
With the new unified config file coming, perhaps we should pivot and support that instead of routes.json.

Also would be great if the emulator can search for the config file, starting from the app_location. Perhaps add an additional command line flag to specify the location of the file, as this info may not be available when starting using something like `--use-app`.

@mkarmark can provide info on the routing, fallback, and response overrides logic in the new config file.

manekinekko:
Notes from today's meeting:
- idea: we could publish the route engine as a binary and do a function call from within the CLI.

PR #11, opened 2020-08-11
Route checks

aaronpowell:
This addresses #6 (but also builds on PR #10 so it's a bit tricky to review).

The implementation works by looking for a `routes.json` file and then loading those roles to test the inbound request first.

I've also done some refactoring to lift up the `users.json` file a bit higher and share it with the proxy to do access control. Long term we might want to think about a better way to have the "current user" handled than a JSON file.


aaronpowell:
@manekinekko - might be easier to look at this PR via https://github.com/aaronpowell/swa-emulator/compare/support-dotnet...aaronpowell:route-checks

bin/index.js
```diff
@@ -75,15 +78,15 @@ const startCommand = [
   `-c 'bgYellow.bold,bgMagenta.bold,bgCyan.bold,bgGreen.bold'`,
 
   // start the reverse proxy
-  `"node ./src/proxy.js"`,
+  `"node ./src/proxy.js --artifacts ${app_artifact_location}"`,
```
manekinekko, line 81: app_artifact_location is available via `SWA_EMU_APP_LOCATION` ([here](https://github.com/manekinekko/swa-emulator/blob/main/bin/index.js#L65))

src/proxy.js
```diff
@@ -20,7 +22,70 @@ const serveStatic = (file, res) => {
...
+  const routesFile = fs.readdirSync(folder).find((file) => file.endsWith("routes.json"));
+
+  if (!routesFile) {
+    return [];
+  }
+
+  return require(path.join(folder, routesFile)).routes || [];
+};
+
+program.option("--artifacts <path>", "path to artifacts").parse(process.argv);
```
manekinekko, line 39: See previous comment about `SWA_EMU_APP_LOCATION `.

src/runtimes.js
```diff
@@ -0,0 +1,19 @@
...
+module.exports.RuntimeType = RuntimeType;
+
+module.exports.detectRuntime = (app_location) => {
+  const files = fs.readdirSync(app_location);
+
+  if (files.some((file) => path.extname(file) === ".csproj")) {
+    return RuntimeType.dotnet;
+  }
+
+  return RuntimeType.node;
```
manekinekko, line 18:
```suggestion
  if (files.some((file) => path.extname(file) === "package.json")) {
    return RuntimeType.node;
  }
  
  return RuntimeType.unknown;
```

src/utils.js
```diff
@@ -112,7 +113,10 @@ module.exports.readConfigFile = () => {
     // these locations must be under the user's project folder
     app_location: path.join(process.cwd(), app_location),
     api_location: path.join(process.cwd(), api_location),
-    app_artifact_location: path.join(process.cwd(), app_location, app_artifact_location),
+    app_artifact_location:
+      detectRuntime(path.join(process.cwd(), app_location)) === RuntimeType.node
+        ? path.join(process.cwd(), app_location, app_artifact_location)
+        : path.join(process.cwd(), app_location, "bin", "Debug", "netstandard2.1", "publish", app_artifact_location),
```
manekinekko, line 119: We can maybe extract this code into a separate function. I can see this logic get complicated once we will have to support more technologies (python, java...).

aaronpowell: Not sure if that'll be the case, remember that this is the front-end runtime, so I don't know what other front-ends we'd support. dotnet is a bit of a special case since there's a WASM story, but we'd have to find out if there were any other supported WASM front-ends (Java to WASM, Python to WASM, etc.) that oryx will understand.

src/runtimes.js
aaronpowell: This is part of the changes in #10 (but this PR builds on top of it), do you want me to close #10 and just have one PR covering a few features?

manekinekko: okay. let's do that. the project is still in alpha. let's go yolo ^^

src/utils.js
manekinekko: My bad 🤦

PR #19, opened 2020-10-01
feat: use existing API dev server

### LINKED ISSUE #17: Providing app/api uri should prevent the emulator from starting its own
Let's say I want to run the APP and/or API myself (maybe they run a compiler like TypeScript), I would run the emulator like so:

```
$> swa --app-uri=http://localhost:3000 --api-uri=http://localhost:7071
```

But when it starts up, it'll attempt to run its own proxies for those, and in particular the API will fail as the port is already in use.

Since there's no need for these proxies to be run by the emulator, it should skip trying to start them if you provide the uri.

manekinekko:
Closes #17


manekinekko:
@aaronpowell do you have some time to review this change?

Issue #16, opened 2020-09-24
Adding concurrently as a dependency breaks the emulator


aaronpowell:
I've installed the emulator as a local devDependency, but I also have concurrently as a local devDependency. Doing this results in the emulator failing to start as it's unable to find the concurrently executable.

It'd probably be better to have concurrently as a peerDependency and use `npx` to run it, rather than directly using it.

manekinekko:
Interesting. What would be the use case for installing the emulator as a local devDep, instead of global CLI?

aaronpowell:
If you're building an app the always needs authentication, then you'd likely want to have it in the npm start.

Also, I have a strong opinion that you should never have global tools, as it's opaque that you need them and causes problems if you need multiple versions installed at once.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

manekinekko:
I am also thinking what would happen for non-Node.js folks–who don't depend on npm/Node.js. They still should be able to get the CLI globally. Any thought?

aaronpowell:
Yes, it should still work globally, mostly I don't understand why concurrently results in an error like this.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

manekinekko:
Would you mind sharing the error message?

manekinekko:
I discussed this issue with Aaron on Teams. When installing SWA locally (devDep), the proxy and concurrently binaries, can't be resolved correctly. 

Currently, `swa` assumes that these binaries are located under `./node_modules` which is not correct when `swa` is installed locally. In this case, the binaries are located in the parent `node_modules`.

Another possible option is to use `npx` to start the binaries.

PR #24, opened 2020-10-19
moving from expected local deps to npx resolved deps

aaronpowell:
This will address #16


manekinekko:
@aaronpowell can you rebasing your changes on main?

aaronpowell:
Closing because it's pretty stale and i'll see if it's still a problem with the recent changes

PR #28, opened 2020-10-31
Porting to TypeScript

manekinekko:
Refactoring https://github.com/manekinekko/swa-emulator/pull/25

Closes https://github.com/manekinekko/swa-emulator/issues/22


patch
```diff
@@ -0,0 +1,289 @@
+From 24edd347438c927ae9e694df5f2c7ab239e7961f Mon Sep 17 00:00:00 2001
```
aaronpowell, line 1: Probably want to delete this file

src/dashboard.ts
```diff
@@ -0,0 +1,117 @@
+// @ts-check
```
aaronpowell, line 1: Does this need to be there? Did I put that in?

manekinekko (commented):
LGTM

PR #31, opened 2020-11-05
Packaging

aaronpowell:
This generates npm packages when a build runs. It'll first attach it as an artifact to the build job (so you could download and install it from the file) but it also pushes to GitHub Packages, so you can install it from there.

The package is versioned using the current semver of the `package.json` but then adds a prerelease of `-ci-<BuildId>` so it sits outside of the standard semver process being used.


aaronpowell:
@manekinekko this helps speed up the dev cycle. I've found that it can be really painful to run from source, but now it can be easily installed

manekinekko (approved):

PR #35, opened 2020-12-04
feat: start the emulator even if a GA isn't found

### LINKED ISSUE #15: Start without the github actions workflow file
Currently, you are required to have created a GitHub Actions workflow file before you're able to use the emulator.

This is because the emulator reads a bunch of stuff out of the workflow file, but it can make for a challenging development flow as you have to have deployed the application before you're able to use the emulator.

It'd be useful if there was a way that you could run the emulator without creating the workflow first.

manekinekko:
Closes #15


manekinekko:
Tests are all passing locally, but somehow they're failing on CI 😓

image

PR #41, opened 2020-12-08
feat: add a start command

### LINKED ISSUE #26: Discussion - Startup commands
```bash
# serve static assets from disk at ./app_artifacts, no api
swa start app_artifacts

# proxy to (ng/react/vue etc) dev server that is already running, no api
swa start http://localhost:4200

# run with api project in ./api
swa start app_artifacts --api api

# run with api project in ./api, but serve api from http://localhost:<port>/functions
swa start app_artifacts --api api --api-prefix functions

# build app with default command (npm run build?) from ./app
# then serve as static assets from build output (./app_artifacts)
swa start app_artifacts --build-app app

# build app with specific command
swa start Client/bin/Debug/net5.0/publish/wwwroot --build-app Client --build-app-command "dotnet publish"

# run with settings in swa.config
swa start 
```

By default, it can try to search for `routes.json` or the new config file and use it.
1. Use the file in the app artifacts folder (if supplied)
1. Search recursively from current folder and use the first config file that it finds.

Optionally, allow specifying location of the routes/config file:

```bash
swa start http://localhost:5000 --config Client/wwwroot/routes.json
```

How VS Code might start the emulator:

```bash
swa start http://localhost:4200 --api api
```

manekinekko:
Implemented user stories:

```
swa start
swa start app_artifacts
swa start http://localhost:4200
swa start app_artifacts --api api
swa start app_artifacts --api api --api-prefix functions
```

Closes #26


anthonychu (changes requested):
Suggested some changes with #60 

Also, wasn't able to pass an absolute path as the start context. For example:

```bash
swa start /Users/antchu/source/ms/swa-cli-js-sample/static
```

I don't think many people will want to do this, so no need to address unless it's an easy fix.

manekinekko:
Will check that out asap. Should not be difficult to fix.

manekinekko:
@anthonychu I just tested your use case and it worked for me without any fix:

```
swa start /Users/wassimchegham/oss/@manekinekko/mslearn-staticwebapp-api/angular-app/dist/angular-app
```

The static server was able to pick up the right location 
image

By the way, the `start` command is expecting an `appArtifactLocation` or a folder that can be served (has an `index.html`). What's inside `/Users/antchu/source/ms/swa-cli-js-sample/static` ?

anthonychu:
Super weird that I'm getting a different result.

image

I think the difference might be that I don't have a workflow file?

manekinekko:
Okay. Will investigate the issue.

manekinekko:
@anthonychu the issue should be fixed by now.

aaronpowell:
First pain point I've had with this is that if I'm running the API already, there's no way to proxy it.

A use case for this is when I'm working on a TypeScript Functions project, I'll run `npm start` which will start `tsc --watch` and the functions tooling, meaning that the Functions are already running.

Can we make it possible to provide a URL for the Functions as well? Or is there another way to tackle it with the `start` command?

aaronpowell:
Oh wait, I see if you were to pass `--api-uri http://localhost:7071` it then uses that

anthonychu:
> Oh wait, I see if you were to pass `--api-uri http://localhost:7071` it then uses that

Wonder if it should just be `--api` to mirror the app location which autodetects if it's a path or uri.

I also don't think we should allow changing the api-prefix right now because the service doesn't support this. Don't want to set the expectation that this is possible when it isn't (yet).

manekinekko:
@anthonychu  I created an issue to track that #64 

@aaronpowell do you think we need to simplify the list of options we pass to the CLI?

aaronpowell:
Yes, I think we need a simplified set of flags, but we should also think through the scenarios of how we expect someone to use it.

Do we expect that people want the cli to run everything for them? To use existing servers? Combination of both?

manekinekko:
For the initial release, I think it will help with adoption if the CLI allows users to:
- starts all necessary servers (the "it just works" scenario)
- integrate with local dev servers (app and api)

PR #60, opened 2021-01-31
Fix startup validation and tests

anthonychu:
Fixes some issues that prevented #41 from working on my machine/setup.

I have a create-react-app app running, listening to localhost:3000. However, `isPortAvailable()` was still returning `true`. 😕 Changed the code to test using a TCP connection instead.

Fixed some tests that were failing because they depended on the name of the folder. They no longer work after the repo was renamed.

### DIFF
--- a/src/cli/commands/start.ts
+++ b/src/cli/commands/start.ts
@@ -6,19 +6,18 @@ import shell from "shelljs";
 import builder from "../../builder";
 import { Dashboard } from "../../dashboard";
 import { createRuntimeHost } from "../../runtimeHost";
-import { GithubActionSWAConfig, isPortAvailable, parseUrl, readConfigFile } from "../../utils";
+import { GithubActionSWAConfig, isAcceptingTcpConnections, isHttpUrl, isPortAvailable, parseUrl, readConfigFile } from "../../utils";
 import { DEFAULT_CONFIG } from "../config";
 
 export async function start(startContext: string, program: CommanderStatic) {
-  if (startContext.startsWith("http")) {
+  if (isHttpUrl(startContext)) {
     // start the emulator and proxy app to the provided uri
-    let { host, port } = parseUrl(startContext);
-    port = port || 4280;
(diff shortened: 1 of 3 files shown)


src/cli/commands/start.ts
```diff
@@ -6,19 +6,18 @@ import shell from "shelljs";
 import builder from "../../builder";
 import { Dashboard } from "../../dashboard";
 import { createRuntimeHost } from "../../runtimeHost";
-import { GithubActionSWAConfig, isPortAvailable, parseUrl, readConfigFile } from "../../utils";
+import { GithubActionSWAConfig, isAcceptingTcpConnections, isHttpUrl, isPortAvailable, parseUrl, readConfigFile } from "../../utils";
 import { DEFAULT_CONFIG } from "../config";
 
 export async function start(startContext: string, program: CommanderStatic) {
-  if (startContext.startsWith("http")) {
+  if (isHttpUrl(startContext)) {
```
anthonychu, line 13: Just in case it's a folder that starts with `http`.

src/cli/commands/start.ts
```diff
@@ -6,19 +6,18 @@ import shell from "shelljs";
...
     // start the emulator and proxy app to the provided uri
-    let { host, port } = parseUrl(startContext);
-    port = port || 4280;
+    let { hostname, port } = parseUrl(startContext);
 
     // make sure host and port are available
     try {
-      const portAvailable = await isPortAvailable({ port, host });
-      if (portAvailable) {
+      const appListening = await isAcceptingTcpConnections({ port, host: hostname });
```
anthonychu, line 19: `isPortAvailable()` was not doing the right thing on my machine. It returned `true` even though my create-react-app dev server was already running. Not sure why. I had to make this change to continue testing.

src/cli/commands/start.ts
manekinekko: Good catch!

PR #69, opened 2021-02-07
fix: check if func binaries exist before starting

### LINKED ISSUE #38: Check if Azure Functions Core Tools is available before starting the api
When the swa cli tries to star the api project using the `func` cli (Azure Functions Core Tools). If `func` is not available, this would trigger an error, an the api won't be emulated.

Check if the `func` cli is already installed. If not, show a warning message or abort the whole process.

**Additional context**
This should work across all supported platforms.

cc @anthonychu

manekinekko:
Closes #38

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -121,12 +121,12 @@ function validateUserConfig(userConfig: Partial<GithubActionSWAConfig>) {
 }
 
 export const readConfigFile = ({ userConfig }: { userConfig?: Partial<GithubActionSWAConfig> } = {}): Partial<GithubActionSWAConfig> | undefined => {
-  const warningMessage = `INFO: Azure Static Web Apps configuration not found under ".github/workflows/"`;
+  const infoMessage = `INFO: GitHub Actions configuration was not found under ".github/workflows/"`;
   const githubActionFolder = path.resolve(process.cwd(), ".github/workflows/");
 
   // does the config folder exist?
   if (fs.existsSync(githubActionFolder) === false) {
-    console.warn(warningMessage);
+    console.info(infoMessage);
     return userConfig && validateUserConfig(userConfig);
   }
 
@@ -139,7 +139,7 @@ export const readConfigFile = ({ userConfig }: { userConfig?: Partial<GithubActi
 
   // does the config file exist?
   if (!githubActionFile || fs.existsSync(githubActionFile)) {
-    console.warn(warningMessage);
+    console.info(infoMessage);
     return userConfig && validateUserConfig(userConfig);
   }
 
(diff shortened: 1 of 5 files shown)


manekinekko:
@aaronpowell would you be able to validate this PR on a Windows machine? 🙏

aaronpowell:
WSL count 😝

I'll do it today

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

src/utils.ts
```diff
@@ -380,18 +380,31 @@ export function computeAppLocationFromArtifactLocation(appArtifactLocation: stri
...
   if (binDirErr) {
-    console.error({ binDirErr });
+    console.error(`ERRR: `, binDirErr);
+    process.exit(-1);
+  }
+  return binDirOutput.stdout.toString().trim();
+}
+
+export function getBinaryPath(binary: string) {
+  const binDirOut = exec(isWindows() ? "npm.cmd" : "npm", ["bin"]);
```
aaronpowell, line 394: You don't need to search for `npm.cmd` on Windows, assuming that `npm` is correctly in the path it will take care of it for you

aaronpowell (changes requested):
Trying to traverse the file system across different OS/shell combinations is going to be fraught with issues (for example, the current Windows detection won't work in PowerShell).

Is there a reason why you wouldn't just use `npx` to execute the commands? That's already got the file system traversal configured so you don't need to do it yourself, and it'll fall back to installing the missing tools as well.

manekinekko:
Thank you @aaronpowell.

For `npx`, I've tried different combinations of options, but each time, `npx` was causing some issues:
- it tried to always download `concurrently` and `http-server` (adding 10+ seconds of starting time when running `swa start ...`).
- curiously, when using flags with `npx --quite`, was causing [these lines](https://github.com/Azure/static-web-apps-cli/blob/main/src/cli/commands/start.ts#L115-L118) to break!

aaronpowell:
weird, with #24 I didn't experience any of those problems. Did you try `--no-install` on `npx` to see what will happen?

manekinekko:
Yes, it still had to install the deps!  Would you mind giving it another try and report your results?

`npx --quite --no-install http-server`

aaronpowell:
_sigh_

Looks like on Windows we have a bigger problem, `npx` has a bug in handling command line arguments - https://github.com/npm/npx/issues/43

I think a better bet would be to programmatically use concurrently, rather than running it from `spawn`.

aaronpowell:
I've opened #71 which shows using `concurrently` programmatically and it can start `http-server` on Windows successfully. There's still some work to be done to get the api working, but I'll wait for a review of #71.

manekinekko:
Closing in favor of #71

PR #70, opened 2021-02-08
Enabling package generation

aaronpowell:
Making it so we now generate packages on builds to make it easier to test

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -50,7 +50,6 @@ jobs:
       name: "Publish to GitHub Packages"
       needs: [package]
       runs-on: ubuntu-18.04
-      if: github.repository_owner == 'manekinekko' # && github.token != ''
       steps:
       - name: Upload
         uses: actions/download-artifact@v2
@@ -60,8 +59,8 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
           registry-url: https://npm.pkg.github.com/
-          scope: "@manekinekko"
-      - run: echo "registry=https://npm.pkg.github.com/@manekinekko" >> .npmrc
+          scope: "@azure"
+      - run: echo "registry=https://npm.pkg.github.com/@azure" >> .npmrc
       - run: npm publish $(ls *.tgz)
         env:
           NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}


.github/workflows/node.js.yml
```diff
@@ -60,8 +59,8 @@ jobs:
...
           registry-url: https://npm.pkg.github.com/
-          scope: "@manekinekko"
-      - run: echo "registry=https://npm.pkg.github.com/@manekinekko" >> .npmrc
+          scope: "@azure"
+      - run: echo "registry=https://npm.pkg.github.com/@azure" >> .npmrc
       - run: npm publish $(ls *.tgz)
         env:
-          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
\ No newline at end of file
+          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
manekinekko, line 66: We probably need to regenerate a new token. I'll check if I have permissions.

aaronpowell: Not needed, that token is provided by GitHub Actions automagically

manekinekko: 😃

manekinekko (approved):

manekinekko:
That's what I was thinking, it seems we can't publish the package that easily !
image

aaronpowell:
I think the reason that the job failed is because _I_ don't have contribute access to the project so it's come from an external source, so the `GITHUB_TOKEN` isn't set to one that has the right scopes for the Azure org.

manekinekko:
Pinging @anthonychu as I think he is working on a similar feature.

anthonychu:
I’m working on publishing to npm. If this is publishing ci packages to GitHub registry I think that’s different.

aaronpowell:
@anthonychu this is publishing to GitHub Packages, ala https://www.aaron-powell.com/posts/2020-11-06-deploy-to-github-packages-with-github-actions/

TL;DR - on each push it'll publish a pre-release version to GitHub Packages. But I'm not sure what impact that'd have with ESRP

manekinekko:
@anthonychu do we want to publish to Github Packages as well? I am ok with either.

manekinekko:
@aaronpowell @anthonychu what's the status of this PR?

aaronpowell:
I'll defer to @anthonychu, but my gut feeling is that the signing process may pose a problem, as we'd have to get the package signed before it could go to GitHub Packages (and maybe even as an artifact of the build) since someone could download and use it.

manekinekko:
I am closing this PR of now. We will open it if needed.

PR #71, opened 2021-02-09
Using concurrently programmatically

aaronpowell:
@manekinekko this is a quick cut of using `concurrently` programmatically rather than trying to spawn the process.

When testing this, it was able to stand up the web server from `npx swa start ./build` on both Windows and WSL. I haven't tried a global install yet though.


manekinekko:
@aaronpowell I am getting these errors when running this branch.
image


My guess is that, when we run `swa` globally, `concurrently` is trying to execute `http-server` from within the user's project. I think we should still need to fetch http-server and other binary from the folder where `swa` is installed.

I like though the fact that we can use `concurrently` programmatically. I didn't know that. Good find!

aaronpowell:
So when you ran it in the above, it was globally installed, but you were using `npx` still?

manekinekko:
Yep, it's a global install. I am using my local dev version of `swa` which is linked to my source `npm link`. 
I tried both `npx swa ...` and `swa ...`. Both gave the same error.

aaronpowell:
@manekinekko I think we'll still have to run `npx` for `http-server`. I just tested it out as both a local and global install and it seems to work (Windows and WSL tested).

Still need to test `build` and the API won't start on Windows as the test for the path is a bash test, which Windows doesn't like 😝

aaronpowell:
Introduced the `build` option back into `start` and I've changed how we look for the `api` folder to determine start there to support Windows.

Seems pretty good on Windows and WSL now.

aaronpowell:
@manekinekko this is ready for review and testing

anthonychu:
@aaronpowell I can test this. Can you explain more about this bit?

> Introduced the `build` option back into `start` and I've changed how we look for the `api` folder to determine start there to support Windows.

aaronpowell:
> @aaronpowell I can test this. Can you explain more about this bit?
> 
> > Introduced the `build` option back into `start` and I've changed how we look for the `api` folder to determine start there to support Windows.

You can get it to do a build before starting, but adding `--build` to the command:

```bash
$> swa start --build
```

For the API, it was originally using a bit of bash to detect the `api` folder (`[ -d <path to api> ]`) which doesn't work on Windows under PowerShell or cmd, so I refactored it to use Node APIs to test for the path.

anthonychu:
I'm not able to run when globally installed:

image

image

aaronpowell:
how did you install it?

anthonychu:
Found the problem. Pushed a quick fix.

aaronpowell:
Oh, did the files move around? Bad merge on my part 😅

manekinekko (approved):

PR #86, opened 2021-02-17
feat: implement the routes engine

### LINKED ISSUE #6: Emulate routes config

manekinekko:
Implementation of the [routes engine](https://docs.microsoft.com/en-us/azure/static-web-apps/configuration):

- [x] headers
- [x] methods
- [x] allowedRoles
- [x] statusCode
- [x] rewrite
- [x] redirect
- [x] navigationFallback
- [x] responseOverrides
- [x] globalHeaders
- [x] mimeTypes

Closes #6 #7


src/proxy/routes-engine.ts
```diff
@@ -0,0 +1,90 @@
...
+      }
+    }
+
+    // rewrite
+    if (userDefinedRoute.rewrite) {
+      req.url = userDefinedRoute.rewrite;
+    }
+
+    // redirect route
+    if (userDefinedRoute.serve || userDefinedRoute.redirect) {
```
manekinekko, line 77: Confirm that the following rule translates to a redirection header?

```
{
            "route": "/logout",
            "redirect": "/.auth/logout"
},
```

src/proxy/routes-engine.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  if (userDefinedRoute) {
+    // set headers
+    if (userDefinedRoute.headers) {
+      for (const header in userDefinedRoute.headers) {
+        res.setHeader(header, userDefinedRoute.headers[header]);
+      }
+    }
+
+    // check allowed method
```
anthonychu, line 44: Method is part of the matching criteria. If the url path matches but there are methods defined, and no methods match, the route is not a match.

src/proxy/routes-engine.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (userDefinedRoute.methods?.includes(req.method as string) === false) {
+      res.writeHead(405);
+      res.end();
+      return;
+    }
+
+    // ACL
+    if (userDefinedRoute.allowedRoles && req.headers.cookie) {
```
anthonychu, line 52: If there are allowed roles, we should perform the auth check whether the cookie is set or not.

src/proxy/routes-engine.ts
```diff
@@ -0,0 +1,90 @@
...
+      const user = decodeCookie(req.headers.cookie);
+      console.log({ user });
+      if (!userDefinedRoute.allowedRoles.some((role) => user?.userRoles.some((ur: string) => ur === role))) {
+        return Promise.resolve(401);
```
anthonychu, line 56: Should this be 403?

src/proxy/routes-engine.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    // redirect route
+    if (userDefinedRoute.serve || userDefinedRoute.redirect) {
+      let route = (userDefinedRoute.serve || userDefinedRoute.redirect) as string;
+      if (route.startsWith(".auth")) {
+        route = `app/${route}`;
+      }
+      // temporarily redirect
+      res.writeHead(302, {
```
anthonychu, line 83: Redirect can have status code specified (301 and 302 are valid, maybe 307 too).

src/proxy/server.ts
```diff
@@ -70,150 +64,107 @@ const injectClientPrincipalCookies = (cookie: string | undefined, proxyReqOrRes:
...
+      });
+      proxyAuth.on("proxyRes", function (proxyRes, req) {
+        console.log("auth>>", req.method, target + req.url);
+        console.log(JSON.stringify(proxyRes.headers, undefined, 2));
+      });
+      proxyAuth.on("error", function (err, req) {
+        console.log("auth>>", req.method, target + req.url);
+        console.log(err.message);
+        proxyAuth.close();
+      });
```
anthonychu, line 109: I've found that setting up handlers in here can be problematic #87

src/proxy/server.ts
```diff
@@ -70,150 +64,107 @@ const injectClientPrincipalCookies = (cookie: string | undefined, proxyReqOrRes:
...
-  }
+    // proxy API request to local API
+    else if (req.url?.startsWith(`/${DEFAULT_CONFIG.apiPrefix}`)) {
+      const target = SWA_CLI_API_URI;
+      console.log("api>", req.method, target + req.url);
+
+      proxyApi.web(req, res, {
+        target,
+      });
+      proxyApi.on("error", function (err, req) {
```
anthonychu, line 120: There are occasional crashes here. See #87 for fix.

anthonychu (commented):
Haven't had a chance to test this out but I've left some comments from reading the code.

And a general comment here is that I think the route matching and processing logic needs to be unit-testable.

src/config.ts
```diff
@@ -10,4 +10,5 @@ export const DEFAULT_CONFIG: SWACLIConfig = {
   appArtifactLocation: `.${path.sep}`,
   appBuildCommand: "npm run build --if-present",
   apiBuildCommand: "npm run build --if-present",
+  swaConfigFilePattern: /(routes|staticwebappsconfig)\.json/,
```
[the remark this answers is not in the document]
itpropro: The routes.json traversal algorithm leads to some problems where the wrong routes.json is chosen. E.g. Nuxt holds the internal nuxt-router config in a file called router.json inside of the .nuxt folder. 
I have a monorepo with an `api` folder and a `frontend` folder. When I specify the frontend folder as app-artifact-location, it will always evaluate to the routes.json in the .nuxt folder. It will probably also iterate over the node_modules folder, where there is a high chance of a routes,json to be existent. 

image

I would recommend to avoid recursion and just evaluate the parent level of the app-artifact-location or maybe introduce a routes_location variable like in the Github workflow.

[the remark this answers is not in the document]
manekinekko: Thank you for reporting this behavior! I will add more check to the traversal algorithm.

manekinekko:
@itpropro would you mind trying this branch and let's know if everything is working as expected for you?

src/core/utils.ts
```diff
@@ -5,6 +5,7 @@ import path from "path";
 import YAML from "yaml";
 import { DEFAULT_CONFIG } from "../config";
 import { detectRuntime, RuntimeType } from "./runtimes";
+const { readdir, readFile } = require("fs").promises;
```
aaronpowell, line 8: Everywhere else we've used the sync methods for working with `fs`, so why change it here? Also, I know it's more combersome, but I'd still use `import` but just do a second destructuring call, since you can't do nested destructuring on `import`:

```js
import { promises as fsPromises } from "fs";

const { readdir, readFile } = fsPromises;
```

src/proxy/server.ts
```diff
@@ -1,21 +1,24 @@
...
+import { navigationFallback } from "./routes-engine/rules/navigationFallback";
+
 const proxyApp = httpProxy.createProxyServer({ autoRewrite: true });
 const proxyApi = httpProxy.createProxyServer({ autoRewrite: true });
 const proxyAuth = httpProxy.createProxyServer({ autoRewrite: false });
-import { decodeCookie, isHttpUrl, validateCookie } from "../core/utils";
-import { DEFAULT_CONFIG } from "../config";
 
-const buildAdress = (host: string, port: number | string | undefined) => `http://${host}:${port}`;
+const address = (host: string, port: number | string | undefined) => `http://${host}:${port}`;
```
aaronpowell, line 14: What would lead to `port` being `undefined`? This does run the risk of the output being `http://<host>:undefined`

src/proxy/routes-engine/rules/mimeTypes.ts
```diff
@@ -0,0 +1,13 @@
+import http from "http";
+
+// See: https://docs.microsoft.com/en-us/azure/static-web-apps/configuration
+export const mimeTypes = async (req: http.IncomingMessage, res: http.ServerResponse, mimeTypes: SWAConfigFileMimeTypes) => {
+  if (req.url?.includes(".")) {
+    const fileExtentionFromURL = req.url?.split(".").pop();
+    const overrideMimeType = mimeTypes[`.${fileExtentionFromURL}`];
```
anthonychu, line 7:
```suggestion
    const overrideMimeType = mimeTypes?.[`.${fileExtentionFromURL}`];
```

I hit a null ref here.

src/proxy/routes-engine/rules/navigationFallback.ts
```diff
@@ -0,0 +1,87 @@
...
+) => {
+  let originlUrl = req.url;
+
+  // don't process .auth or api requests
+  if (originlUrl?.startsWith("/.auth") || originlUrl?.startsWith("/api")) {
+    return;
+  }
+
+  // exit if no rewrite rule provided
+  if (!navigationFallback.rewrite) {
```
anthonychu, line 22:
```suggestion
  if (!navigationFallback?.rewrite) {
```

I got a null ref here

anthonychu (changes requested):
Ran it on a project (it has routes.json) and got a couple of null reference exceptions. Let's try to make sure we check carefully for places where something can be undefined or null.

src/core/utils.ts
manekinekko: I think that using the sync calls was a mistake from the start. I think we still have time to change that.

src/proxy/server.ts
manekinekko: Same goes for `host`. These values are populated by the user's environment vars, and the default values are set by the CLI flags.

I am gonna add some checks but the user should still be able to override values using their own env vars.

src/proxy/routes-engine/rules/navigationFallback.ts
manekinekko: Do we have the specs for the routes.json config? I'd love to know what property is mandatory and which is not.

manekinekko:
Given the short amount of time we have, I am going to merge this PR. Feel free to open new PR if you think something needs to be updated.

Issue #4, opened 2020-07-07
Support HTTPS locally


manekinekko:
Currently, the CLI serves the traffic over HTTP. We need to support HTTPS.

- [ ] generate a valid SSL certificate for localhost (see [scripts/cert.sh](https://github.com/Azure/static-web-apps-cli/blob/main/scripts/cert.sh))
- [ ] use HTTPS for the proxy

manekinekko:
Fixed in #140

PR #140, opened 2021-03-23
feat: Support serving web and api over https locally

yamachu:
ref: https://github.com/Azure/static-web-apps-cli/issues/4

Some Web API requires contents served over https (for example: Navigator.mediaDevices.getUserMedia).
Since it can be used if it is provided from a specific location such as localhost and file protocol, we can use by using localhost over http, which is currently available.
So it's a low priority because there are alternatives ways, but I guess this feature will bring it closer to production.

I tried the following command

```
$ pwd
~/Projects/github.com/yamachu/static-web-apps-cli
$ SWA_CLI_APP_USE_HTTPS=true SWA_CLI_APP_SSL_KEY="./scripts/localhost.key" SWA_CLI_APP_SSL_CERT="./scripts/localhost.crt" npm start start http://localhost:8000
```

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -35,8 +35,8 @@ import { start } from "./commands/start";
     .option<number>("--port <port>", "set the cli port", parsePort, DEFAULT_CONFIG.port)
     .option("--build", "build the app and API before starting the emulator", false)
     .option("--ssl", "serving the app and API over HTTPS", DEFAULT_CONFIG.ssl)
-    .option("--ssl-cert <sslCertLocation>", "set the location where ssl certificate file is used for serving", DEFAULT_CONFIG.sslCert)
-    .option("--ssl-key <sslKeyLocation>", "set the location where ssl key file is used for serving", DEFAULT_CONFIG.sslKey)
+    .option("--ssl-cert <sslCertLocation>", "SSL certificate to use for serving HTTPS", DEFAULT_CONFIG.sslCert)
+    .option("--ssl-key <sslKeyLocation>", "SSL key to use for serving HTTPS", DEFAULT_CONFIG.sslKey)
     .action(async (context: string = `.${path.sep}`, options: any) => {
       options = {
         ...options,
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -34,6 +34,9 @@ import { start } from "./commands/start";
     .option("--host <host>", "set the cli host address", DEFAULT_CONFIG.host)
     .option<number>("--port <port>", "set the cli port", parsePort, DEFAULT_CONFIG.port)
     .option("--build", "build the app and API before starting the emulator", false)
+    .option("--ssl", "serving the app and API over HTTPS", DEFAULT_CONFIG.ssl)
+    .option("--ssl-cert <sslCertLocation>", "set the location where ssl certificate file is used for serving", DEFAULT_CONFIG.sslCert)
+    .option("--ssl-key <sslKeyLocation>", "set the location where ssl key file is used for serving", DEFAULT_CONFIG.sslKey)
     .action(async (context: string = `.${path.sep}`, options: any) => {
       options = {
         ...options,
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,7 +4,7 @@ export const DEFAULT_CONFIG: SWACLIConfig = {
   host: "0.0.0.0",
   apiPort: 7071,
   apiPrefix: "api",
-  useHttps: "false",
+  useHttps: false,
   appLocation: `.${path.sep}`,
   appArtifactLocation: `.${path.sep}`,
   appBuildCommand: "npm run build --if-present",
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,7 +4,7 @@ export const DEFAULT_CONFIG: SWACLIConfig = {
   host: "0.0.0.0",
   apiPort: 7071,
   apiPrefix: "api",
-  useHttps: false,
+  ssl: false,
   appLocation: `.${path.sep}`,
   appArtifactLocation: `.${path.sep}`,
   appBuildCommand: "npm run build --if-present",
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,6 +4,7 @@ export const DEFAULT_CONFIG: SWACLIConfig = {
   host: "0.0.0.0",
   apiPort: 7071,
   apiPrefix: "api",
+  useHttps: "false",
   appLocation: `.${path.sep}`,
   appArtifactLocation: `.${path.sep}`,
   appBuildCommand: "npm run build --if-present",
--- a/src/config.ts
+++ b/src/config.ts
@@ -7,6 +7,8 @@ export const DEFAULT_CONFIG: SWACLIConfig = {
   ssl: false,
   appLocation: `.${path.sep}`,
   appArtifactLocation: `.${path.sep}`,
+  sslCert: undefined,
+  sslKey: undefined,
   appBuildCommand: "npm run build --if-present",
   apiBuildCommand: "npm run build --if-present",
   swaConfigFilename: "staticwebapp.config.json",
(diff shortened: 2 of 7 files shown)


manekinekko:
Thank you for your contribution @yamachu 🎉

src/core/utils.test.ts
```diff
@@ -770,5 +770,10 @@ jobs:
       expect(address("127.0.0.1", "4200")).toBe("http://127.0.0.1:4200");
       expect(address("[::1]", "4200")).toBe("http://[::1]:4200");
     });
+
+    it("should accept protocol", () => {
```
manekinekko, line 774:
```suggestion
    it("should accept both HTTP and HTTPS protocols", () => {
```

src/config.ts
```diff
@@ -4,6 +4,7 @@ export const DEFAULT_CONFIG: SWACLIConfig = {
   host: "0.0.0.0",
   apiPort: 7071,
   apiPrefix: "api",
+  useHttps: "false",
```
manekinekko, line 7:
```suggestion
  useHttps: false,
```

src/proxy/server.ts
```diff
@@ -18,8 +19,11 @@ const SWA_CLI_PORT = parseInt((process.env.SWA_CLI_PORT || DEFAULT_CONFIG.port)
 const SWA_CLI_API_URI = address(SWA_CLI_HOST, process.env.SWA_CLI_API_PORT);
 const SWA_CLI_APP_LOCATION = (process.env.SWA_CLI_APP_LOCATION || DEFAULT_CONFIG.appLocation) as string;
 const SWA_CLI_APP_ARTIFACT_LOCATION = (process.env.SWA_CLI_APP_ARTIFACT_LOCATION || DEFAULT_CONFIG.appArtifactLocation) as string;
+const SWA_CLI_APP_USE_HTTPS = (process.env.SWA_CLI_APP_USE_HTTPS || DEFAULT_CONFIG.useHttps) === "true";
```
manekinekko, line 22:
```suggestion
const SWA_CLI_APP_USE_HTTPS = process.env.SWA_CLI_APP_USE_HTTPS === "true" || DEFAULT_CONFIG.useHttps === true;
```

src/proxy/server.ts
```diff
@@ -34,14 +38,25 @@ if (SWA_WORKFLOW_CONFIG_FILE) {
   logger.info(`\nFound workflow file:\n    ${chalk.green(SWA_WORKFLOW_CONFIG_FILE)}`);
 }
 
+if (SWA_CLI_APP_USE_HTTPS && (SWA_CLI_APP_SSL_KEY === undefined || SWA_CLI_APP_SSL_CERT === undefined)) {
+  logger.error(`If use HTTPS, must be set SWA_CLI_APP_SSL_KEY and SWA_CLI_APP_SSL_CERT`, true);
```
manekinekko, line 42:
```suggestion
  logger.error(`SSL Key or SSL Cert are required when using HTTPS`, true);
```

src/swa.d.ts
```diff
@@ -56,6 +56,7 @@ declare type SWACLIConfig = GithubActionWorkflow & {
   port?: number;
   host?: string;
   apiPort?: number;
+  useHttps?: "true" | "false" | string;
```
manekinekko, line 59:
```suggestion
  useHttps?: boolean;
```

manekinekko:
@yamachu I left some comments. Also, we would like to enable HTTPS with a CLI flag, like `--ssl`. 

Here are some options:
```
swa start ./public \
    --ssl true \
    --ssl-cert "/home/user/ssl/example.crt" \
    --ssl-key "/home/user/ssl/example.key"
```

Or by simply passing `--ssl`, in this case we will read Key and Cert from a default folder, for example called `./ssl`:
```
swa start ./public --ssl
```

Do you think you can update this PR with these changes?

Cheers.

yamachu:
@manekinekko Thanks for reviewing!
I've completed the correction of the suggested codes.

There's one thing I didn't make sense. 

> Or by simply passing --ssl, in this case we will read Key and Cert from a default folder, for example called ./ssl:

https://github.com/Azure/static-web-apps-cli/pull/140/commits/20db8d014eeb33f7a4898612b2b1f82900ef34a3#diff-56c6d11b6169bded929472469c7a06cf11c7fe4ec6905aac0199026edb3e4a48R92-R93

Did you intend to use `copy-assets` and load what is included in this application (localhost.crt, localhost.key)?

manekinekko:
Thank you @yamachu 

We haven't designed this yet. So I was actually suggesting that we read a folder `./ssl` from the user root application folder (`--app-location`).
```
.
├── package.json
├── app/
├── api/
├── ssl/**/*.{key, cert} 
```

Do you think this is a valid use case? Do developers usually save their Key and Cert on GitHub?

yamachu:
@manekinekko 

> we read a folder ./ssl from the user root application folder (--app-location).

I guess that this is a valid use case and looks good to me.

```
.
├── package.json
├── app/
├── api/
├── proxy/dev_ssl_certs/*.{key, cert}
```

This is my projects structure that I often use when serving it by proxy application such as nginx.

> Do developers usually save their Key and Cert on GitHub?

As far as I know, no one else is doing that.
In most cases, there is (maybe) no issue if we commit it and push, because we set `localhost` to SAN.
However, I haven't committed because I don't want to admit exceptions in terms of broken window theory in my case.

manekinekko:
Thank you @yamachu 

Let's then just add support for CLI flags, like following:

```
swa start ./public \
    --ssl \
    --ssl-cert "/home/user/ssl/example.crt" \
    --ssl-key "/home/user/ssl/example.key"
```

yamachu:
Thanks @manekinekko 

I modified it to the CLI options you suggested.
I wasn't sure if the `commander` could support a combination of CLI options, so I used an if statement.
https://github.com/Azure/static-web-apps-cli/pull/140/commits/8da34925521ae0411667fe7a041105622b1e7c3b#diff-56c6d11b6169bded929472469c7a06cf11c7fe4ec6905aac0199026edb3e4a48R87-R88
Is that ok?

src/cli/index.ts
```diff
@@ -34,6 +34,9 @@ import { start } from "./commands/start";
     .option("--host <host>", "set the cli host address", DEFAULT_CONFIG.host)
     .option<number>("--port <port>", "set the cli port", parsePort, DEFAULT_CONFIG.port)
     .option("--build", "build the app and API before starting the emulator", false)
+    .option("--ssl", "serving the app and API over HTTPS", DEFAULT_CONFIG.ssl)
+    .option("--ssl-cert <sslCertLocation>", "set the location where ssl certificate file is used for serving", DEFAULT_CONFIG.sslCert)
```
manekinekko, line 38:
```suggestion
    .option("--ssl-cert <sslCertLocation>", "SSL certificate to use for serving HTTPS", DEFAULT_CONFIG.sslCert)
```

src/cli/index.ts
```diff
@@ -34,6 +34,9 @@ import { start } from "./commands/start";
...
+    .option("--ssl-key <sslKeyLocation>", "set the location where ssl key file is used for serving", DEFAULT_CONFIG.sslKey)
```
manekinekko, line 39:
```suggestion
    .option("--ssl-key <sslKeyLocation>", "SSL key to use for serving HTTPS", DEFAULT_CONFIG.sslKey)
```

manekinekko:
That's perfect. One last minor change and your PR will be ready to be merged.

manekinekko (approved):

manekinekko (approved):

manekinekko:
Congratulations @yamachu 🎉

iMicknl:
@manekinekko @yamachu are there plans to support self signed cert generation, similar to what dotnet offers? Or at least integrate with a tool that makes it easy to use. Currently it is quite a hassle to set up, especially when you share the repository with multiple members.

Should I log this as a separate feature request?

https://www.hanselman.com/blog/developing-locally-with-aspnet-core-under-https-ssl-and-selfsigned-certs

anthonychu:
Yes please log a new issue to track this feature request.

Issue #112, opened 2021-03-12
CTRL-C took 2 minutes to apply


johnpapa:
**Describe the bug**
CTRL-C many times does not close the running process, for about 2 minutes

**To Reproduce**
1. Using the shopathome monorepo, run the `swa start` in the root
2. CTRL-C many times
3. 

**Expected behavior**
Should kill the process right away

**Additional context**
![image](https://user-images.githubusercontent.com/1202528/110968679-d37ac880-8325-11eb-9ec2-f28a776443bd.png)

manekinekko:
Interesting! I am unable to reproduce this. 
![image](https://user-images.githubusercontent.com/1699357/110975111-97fcf000-835f-11eb-870b-737b30e80b0c.png)

I see a GET request there, have you tried to access the app from a browser before killing the process?

johnpapa:
I did refresh web page once.

anthonychu:
I've run into this on macOS a couple of times.

PR #151, opened 2021-04-02
fix: force exit on SIGINT/SIGTERM

anthonychu:
fixes: #112


anthonychu:
@johnpapa See if this addresses your issue as well.

Issue #178, opened 2021-04-09
404 pages will not be displayed with fallback routes in config


k-miyake:
**Describe the bug**
Vue.js app are served with 0.3.0 CLI.
When put the fallback routes in config, 404 page is not displayed.

Defined fallback routes following in ` staticwebapp.config.json`.

```json
{
    "navigationFallback": {
        "rewrite": "/index.html",
        "exclude": ["/js/*.{js,map}", "/img/*.{png,jpg,gif}", "/css/*"]
    }
}
```

**To Reproduce**
1. Build the Vue application with `npm run build`
2. Run `swa start ./dist`
3. Enter the URL of a path that does not exist, for example, `http://localhost:4280/abc`.
4. You will see a screen with nothing rendered.

**Expected behavior**
404 page will be displayed.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 89.0.4389.114

**Additional context**
SWA CLI is 0.3.0 installed via npm

manekinekko:
Thank you @k-miyake for creating this issue. it looks like this is a duplicate or related to #168 that I am already working on. Will keep you posted.

manekinekko:
@k-miyake I tried your steps on a local Vue app and I can see my custom message. Can you share a screenshot of your issue?

![image](https://user-images.githubusercontent.com/1699357/114152796-f7c6c680-991e-11eb-8c3c-da549883840c.png)

manekinekko:
btw, in my Vue router configuration, I've setup a custom 404 component:

```

export default new Router({
  mode: 'history',
  base: process.env.BASE_URL,
  routes: [
   ...
    {
      path: '*',
      component: PageNotFound,
    },
  ],
});

```

k-miyake:
@manekinekko 
My explanation may have been bad.
I wanted to know how to get the default 404 page of the SWA CLI to show up, not the custom 404 page not showing up.
If you remove the fallback setting, the SWA CLI's 404 page will be displayed.
![404](https://user-images.githubusercontent.com/1104403/114157850-092fc300-995f-11eb-8d65-bb9c35147605.png)

k-miyake:
@manekinekko 
I realized that since it's an SPA, it's impossible to display the SWA 404 page as long as I set a fallback.
I'll wait for the #168 fix and use custom 404.
Thanks.

manekinekko:
[...]

PR #179, opened 2021-04-09
fix: allow users to set custom error pages

### LINKED ISSUE #168: Show 404 page if can't serve a file
This is a regression bug. We used to show a custom [404 page](https://github.com/Azure/static-web-apps-cli/blob/main/src/public/404.html), but this seems to be broken now. The logic responsible for this should be here:

https://github.com/Azure/static-web-apps-cli/blob/cf9acb6e06d4e152c6ad95ce9827719b11bb5f1b/src/proxy/server.ts#L206-L231

manekinekko:
Fixes #168 #178


manekinekko:
e2e are failing because of CI misconfig.

PR #195, opened 2021-05-10
fix: redirect to the url which starts with `https://` when logged out if `swa` is started with `--ssl` flag

horihiro:
Currently, a request to `/.auth/logout` always is redirect to `http://`.
https://github.com/Azure/static-web-apps-cli/blob/db13f3a2646759b5b917733e4eb895674539b345/src/auth/routes/auth_logout.ts#L14

Because this causes a redirection to an invalid URL when `swa` is started with `--ssl` flag, this PR switches the protocol of URL by the value of `process.env.SWA_CLI_APP_SSL`.

### DIFF
--- a/src/auth/routes/auth_logout.ts
+++ b/src/auth/routes/auth_logout.ts
@@ -11,7 +11,7 @@ const httpTrigger = async function (context: Context, req: http.IncomingMessage)
     return;
   }
 
-  const uri = `http${process.env.SWA_CLI_APP_SSL === "true" ? "s" : ""}://${host}`;
+  const uri = `${process.env.SWA_CLI_APP_SSL === "true" ? "https" : "http"}://${host}`;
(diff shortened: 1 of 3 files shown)


src/auth/routes/auth_logout.ts
```diff
@@ -11,7 +11,7 @@ const httpTrigger = async function (context: Context, req: http.IncomingMessage)
     return;
   }
 
-  const uri = `http://${host}`;
+  const uri = `http${process.env.SWA_CLI_APP_SSL === "true" ? "s" : ""}://${host}`;
```
manekinekko, line 14:
```suggestion
  const uri = `${process.env.SWA_CLI_APP_SSL === "true" ? "https" : "http"}://${host}`;
```

manekinekko (changes requested):

manekinekko (approved):

PR #219, opened 2021-06-02
feat: allow --run to take any command

### LINKED ISSUE #147: Allow custom startup commands

When starting the emulator with a dev server, we need to start the dev server separately before starting `swa`.

Allow an app startup command and working folder to be provided. The emulator starts the command before proceeding. For example, running a startup command in the frontend folder might look like this:

```
swa start http://localhost:3000 --app-startup-command "npm start" --app-location ./frontend
```

Or if we can make this work cross platform:

```
swa start http://localhost:3000 --app-startup-command "cd frontend && npm start"
```

Perhaps also support `--api-startup-command`? Not sure if this is needed right now.

Start the process separately.

**Additional context**

Also needs this functionality as we have to wait for the servers to be up before proceeding: #146

anthonychu:
Resolves #147 

Support passing arbitrary commands to `--run`, such as:
* `swa start http://localhost:5000 --run "dotnet watch run"`
* `swa start http://localhost:5000 --run "cd MyApp && dotnet watch run"`

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -70,6 +70,22 @@ To use SWA CLI with your local dev server, follow these two steps:
    swa start http://<APP_DEV_SERVER_HOST>:<APP_DEV_SERVER_PORT>
    ```
 
+Instead of starting a dev server separately, you can provide the startup command to the CLI.
+
+```bash
+# npm start script (React)
+swa start http://localhost:3000 --run "npm start"
+
+# dotnet watch (Blazor)
+swa start http://localhost:5000 --run "dotnet watch run"
+
+# Jekyll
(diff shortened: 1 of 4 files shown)


manekinekko:
LGTM. However, having both the swa cli and the --run command write to the same terminal stdout might lead to some UX issues. Also, some dev servers might also automatically open the browser pointing to the dev server's port (not the saw cli port), which will be confusing for folks.

We probably need to add these warning/disclaimers in the readme. wdyt?

anthonychu:
Good points. Both of them are existing problems with what `--run` supports today. I think we can see what feedback come in and adjust accordingly.

manekinekko (approved):

PR #225, opened 2021-06-03
fix: serve 4xx pages when using dev server

### LINKED ISSUE #224: 4xx pages aren't served properly when using dev server
**Describe the bug**
When a route is blocked by allowRoles, the 4xx page request is proxied to the dev server instead of returning the page with the correct status code.

**To Reproduce**
1. Create route that is blocked:
    ```json
    "routes": [
        {
            "route": "/new",
            "allowedRoles": ["author"]
        }
    ]
    ```
2. Start app with dev server `swa start http://localhost:3000 --api api --run "npm:start"`
3. Navigate to the page.

**Expected behavior**
built in 401 page should be returned

anthonychu:
Fixes #224

### DIFF
--- a/src/msha/middlewares/request.middleware.ts
+++ b/src/msha/middlewares/request.middleware.ts
@@ -107,8 +107,10 @@ function serveStaticOrProxyReponse(req: http.IncomingMessage, res: http.ServerRe
     logger.silly(` - target: ${chalk.yellow(target)}`);
   }
 
+  const is4xx = res.statusCode >= 400 && res.statusCode < 500;
+
   // if the static app is served by a dev server, forward all requests to it.
-  if (IS_APP_DEV_SERVER()) {
+  if (IS_APP_DEV_SERVER() && !is4xx) {
     logger.silly(`remote dev server detected. Proxying request`);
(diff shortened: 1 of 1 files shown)


src/msha/middlewares/request.middleware.ts
```diff
@@ -107,8 +107,10 @@ function serveStaticOrProxyReponse(req: http.IncomingMessage, res: http.ServerRe
     logger.silly(` - target: ${chalk.yellow(target)}`);
   }
 
+  const is4xx = res.statusCode >= 400 && res.statusCode < 500;
+
   // if the static app is served by a dev server, forward all requests to it.
-  if (IS_APP_DEV_SERVER()) {
+  if (IS_APP_DEV_SERVER() && !is4xx) {
```
manekinekko, line 113: I think with this condition, If the user has custom 401/403/404 pages defined (as response overrides), we won't be able to serve them when running in dev server mode.

anthonychu: Looks like it needs both checks. Pushed an update that uses dev server if it's a custom url or it's not 4xx.

manekinekko (approved):

manekinekko:
LGTM.

PR #246, opened 2021-06-18
fix: bind to http://localhost by default

### LINKED ISSUE #242: Show http://localhost:4280 as the main emulator URL
Currently, this is printed on startup:

```
[swa] Available on:
[swa]     http://0.0.0.0:4280
[swa]     http://192.168.187.166:4280
```

Not sure how often those addresses are used. I'd imagine 99% of the time, we want folks to use http://localhost:4280. Consider changing this output to simply:

```
[swa] Available on: http://localhost:4280
```

Or remove it altogether and just add the URL to this message: 

```
Azure Static Web Apps emulator started at http://localhost:4280 . Press CTRL+C to exit.
```

manekinekko:
Fixes #242

image

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,7 +1,7 @@
 import path from "path";
 export const DEFAULT_CONFIG: SWACLIConfig = {
   port: 4280,
-  host: "0.0.0.0",
+  host: "localhost",
   apiPort: 7071,
   apiPrefix: "api",
   ssl: false,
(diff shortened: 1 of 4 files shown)


anthonychu:
I wonder if we should continue to bind to 0.0.0.0 and print localhost. There’s a chance that if running inside a container and it needs to listen to an external port, binding to localhost isn’t enough.

Also, I don’t know if it’s possible but can we write a test to ensure `Azure Static Web Apps emulator started` is always printed when the emulator is ready? The reason for this is VS Code needs a string to watch for to indicate that the emulator is ready.

manekinekko:
I will include a test for the message.

Binding to `0.0.0.0` can still be possible using `--host 0.0.0.0` if the user needs to access the CLI from outside of localhost (container or mobile device...).

PR #249, opened 2021-06-24
Print notice that the emulator may not match the cloud environment

### LINKED ISSUE #230: Print notice that the emulator may not match the cloud environment

There have been some issues filed against the static-web-apps repo because of differences in behavior between the emulator and the Azure environment.

Print a notice on swa start command:

> This CLI runs an emulator that may not match the cloud environment exactly. Always deploy and test your app in Azure.

We’ve added a similar message to the readme but folks might not see it.

Also make it clear that the CLI is still in preview (but the service is GA).

#229

sgollapudi77:
Fixes #230


sgollapudi77:
Screenshot (13)

PR #255, opened 2021-07-08
feat: Configure await timeout before connecting to a dev server

### LINKED ISSUE #253: Allow configuring the await timeout before connecting to a dev server
When running the CLI against a dev server, we currently await 30 seconds (by default) for the dev server to start before connect to it (see https://github.com/Azure/static-web-apps-cli/issues/252)

So users have reported that some dev servers might take longer. We need a new flag to configure that timeout.

Introduce a new flag `--wait-on-timeout` that maps to an env variable `SWA_CLI_WAIT_ON_TIMEOUT`, and update all references to `validateDevServerConfig()` defined in: https://github.com/Azure/static-web-apps-cli/blob/95b6e5641707abaef42b2db7fb21885e8783f5bc/src/core/utils/net.ts#L48

cc @anthonychu

sgollapudi77:
Resolves #253 
Allows user to configure wait timeout before dev server starts using the flag `--devserver-timeout`, such as:
`swa start --devserver-timeout 10000`

### DIFF
--- a/src/core/utils/net.ts
+++ b/src/core/utils/net.ts
@@ -157,12 +157,3 @@ export function hostnameToIpAdress(hostnameOrIpAddress: string | undefined) {
   }
   return hostnameOrIpAddress;
 }
-
-export function parseTime(time: string) {
-  //The argument 10 implies to convert the given string to base-10(decimal)
-  const timeValue = parseInt(time, 10);
-  if (isNaN(timeValue)) {
-    logger.error(`Time "${time} is not a number.`, true);
-  }
-  return timeValue;
-}
--- a/src/core/utils/net.ts
+++ b/src/core/utils/net.ts
@@ -45,7 +45,7 @@ export function isHttpUrl(url: string) {
  * @param timeout Maximum time in ms to wait before exiting with failure (1) code,
   default Infinity.
  */
-export async function validateDevServerConfig(context: string, timeout = 30000) {
+export async function validateDevServerConfig(context: string, timeout: number) {
   let { hostname, port } = parseUrl(context);
 
   try {
@@ -156,4 +156,13 @@ export function hostnameToIpAdress(hostnameOrIpAddress: string | undefined) {
     return "127.0.0.1";
   }
   return hostnameOrIpAddress;
+}
+
+export function parseTime(time: string) {
+  //The argument 10 implies to convert the given string to base-10(decimal)
+  const timeValue = parseInt(time, 10);
+  if (isNaN(timeValue)) {
+    logger.error(`Time "${time} is not a number.`, true);
+  }
+  return timeValue;
 }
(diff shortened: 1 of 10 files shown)


sgollapudi77:
Screenshot (27)

sgollapudi77:
@manekinekko can you please check the PR once and tell me if I need to add some checks (or) do I need to change some naming convention (or) anything at all. 
Thanks

src/core/utils/net.ts
```diff
@@ -157,3 +157,12 @@ export function hostnameToIpAdress(hostnameOrIpAddress: string | undefined) {
   }
   return hostnameOrIpAddress;
 }
+
+export function parseTime(time: string) {
+  //The argument 10 implies to convert the given string to base-10(decimal)
+  const timeValue = parseInt(time, 10);
+  if (isNaN(timeValue)) {
+    logger.error(`Time "${time} is not a number.`, true);
```
manekinekko, line 165:
```suggestion
    logger.error(`--devserser-timeout should be a number expressed in milliseconds. Got "${time}".`, true);
```

src/core/utils/net.ts
```diff
@@ -157,3 +157,12 @@ export function hostnameToIpAdress(hostnameOrIpAddress: string | undefined) {
   }
   return hostnameOrIpAddress;
 }
+
+export function parseTime(time: string) {
```
manekinekko, line 161:
```suggestion
export function parseDevServerTimeout(time: string) {
```

src/core/utils/net.ts
```diff
@@ -157,3 +157,12 @@ export function hostnameToIpAdress(hostnameOrIpAddress: string | undefined) {
   }
   return hostnameOrIpAddress;
 }
+
```
manekinekko, line 160: This function should be moved to `src/core/utils/cli.ts` and must be tested.

manekinekko (changes requested):

manekinekko:
I left a few comments but overall looks great!

sgollapudi77:
> I left a few comments but overall looks great!

src/core/utils/net.ts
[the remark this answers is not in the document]
sgollapudi77: Updated the same

src/core/utils/net.ts
sgollapudi77: Updated the same

src/core/utils/net.ts
sgollapudi77: Updated and added tests

sgollapudi77:
@manekinekko I've made changes but made a mistake of pushing 2 commits to this PR.

manekinekko (approved):

manekinekko:
LGTM

PR #257, opened 2021-07-20
docs: clarify use with dev server

shmool:
I had trouble using the CLI through Angular dev server until I understood that I still need to use the SWA CLI port.


shmool:
Doesn't pass PR checks

Issue #275, opened 2021-08-21
Waiting for API to be ready fails intermittently


codekitchen:
**Describe the bug**

Sometimes when starting up `swa`, it hangs on waiting for the API to be available until it times out and exits. Other times it starts up fine. I've been seeing this intermittently for a while and I believe I've tracked it down to [validateDevServerConfig](https://github.com/Azure/static-web-apps-cli/blob/e6cbf081971b3f9418e197c7d63f9d5aac3ac194/src/core/utils/net.ts#L52).

That validation uses two different techniques -- for the first check, it just opens a raw TCP socket. If that fails, it falls back to the `wait-on` library which sends an actual HTTP GET request and validates the response. Since my API function returns a 400 error code for `GET /` with no parameters, `wait-on` in the current configuration always fails the check. So I think what is happening is `swa` succeeds when my API happens to spin up first and only the raw socket check happens, and fails when swa happens to win the race and the `wait-on` checks happen.

**To Reproduce**
- Create an API function that returns 400 error code on a bare `GET /` request.
- `swa start`
- Depending on if the API starts accepting connections before swa tries to the connect the first time, it'll either succeed or hang forever waiting until it times out and exits.

**Expected behavior**
`swa` shouldn't fail the check if my API function is handling requests but returning an HTTP error response. Ideally also the behavior should be consistent whichever process wins the race.

**Screenshots**

<img width="679" alt="Screen Shot 2021-08-20 at 9 21 11 PM" src="https://user-images.githubusercontent.com/52172/130308890-2924fafd-1145-4581-a548-d5291ce69b82.png">

**Desktop (please complete the following information):**
 - OS:  MacOS Big Sur 11.5.2 (20G95)
 - Browser [e.g. chrome, safari]: N/A

codekitchen:
As a workaround for now, I've added a `sleep 3` to my `concurrently` commands so that my API is always up and running before `swa` tries to connect the first time.

manekinekko:
Hey @codekitchen we have added a new flag in the latest release to allow you to wait for the dev server to start:

`swa start ./ --devserver-timeout 60000`

> I just noticed that this is missing from our docs, will add it asap.

codekitchen:
[...]

PR #281, opened 2021-08-23
fix(core): wait on API using tcp

codekitchen:
Wait on the API to come up by just connecting over TCP rather than making HTTP requests.

Closes Waiting for API to be ready fails intermittently #275

### DIFF
--- a/src/core/utils/net.ts
+++ b/src/core/utils/net.ts
@@ -55,7 +55,7 @@ export async function validateDevServerConfig(context: string, timeout: number) 
       try {
         spinner.start(`Waiting for ${chalk.green(context)} to be ready`);
         await waitOn({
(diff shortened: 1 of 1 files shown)


codekitchen:
Testing note: I confirmed locally that this fixes my project. However, my M1 Macbook doesn't work with `azure-functions-core-tools` so I spin up my API function separately using a shim startup script and pass `--api` to swa. I'm unable to test the more common use case.

sgollapudi77 (commented):
Looks good to me. I believe connecting with TCP instead of HTTP doesn't have any side effects.

manekinekko (approved):

manekinekko:
LGTM

sgollapudi77:
Closing it because the changes seem fine.

Issue #286, opened 2021-08-26
rewrites in development have incorrect path


mcansh:
Before filing this issue, please ensure you're using the latest CLI by running `swa --version` and comparing to the latest version on [npm](https://www.npmjs.com/package/@azure/static-web-apps-cli).

**Are you accessing the CLI from the default port `:4280` ?**
- [ ] No, I am using a different port number (--port) and accessing the CLI from that port
- [x] Yes, I am accessing the CLI from port `:4280`

> Make sure you are accessing the URL printed in the console when running!

**Describe the bug**
Hey! I'm working on bringing Azure support to [Remix](https://remix.run)! I have a rewrite which rewrites all paths to a function and when I start up my local server `swa start ./public --run "remix dev" --api .` and console.log the current path, I'd expect to see it be `/`, but I'm getting that of the function `/api/azure` - here's my app's repo https://github.com/mcansh/remix-on-azure and here it is working when deployed https://brave-wave-0feb78710.azurestaticapps.net

**To Reproduce**
1. Create a function and rewrite all paths to point to that function
2. Type in command `swa start ./public --run "remix dev" --api .`
3. Go to http://localhost:4280
5. See error

**Expected behavior**

our request url will be "/"

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] macOS
 - Browser [e.g. chrome, safari] chrome, safari, firefox
 - Version [e.g. 22] 95.0.4623.0, 15.0, 93.0a1

**Additional context**

danielgary:
I investigated this issue a few months back.  Remix needs the original URL to be preserved even after the fallback.  In production, this works and we can access the original URL using the ```x-ms-original-url``` header, but that header is not set when using the cli.

PR #287, opened 2021-08-26
fix: Using a function as a fallback route doesn't work

### LINKED ISSUE #261: Using a function as a fallback route doesn't work
**Describe the bug**
Functions should be able to be used as navigation fallback like configured [here](https://github.com/anthonychu/test-swa-ssr/blob/758d54b98c26c383a1cb5535c51cd4b03309f877/public/staticwebapp.config.json#L12-L14) but it doesn't work.

**To Reproduce**
1. git clone https://github.com/anthonychu/test-swa-ssr.git
2. cd test-swa-ssr
3. swa start public --api api
4. Navigate to http://localhost:4280/foo returns the 404 page

**Expected behavior**
Navigating to http://localhost:4280/foo should return the function response. This works when deployed to Azure.

sgollapudi77:
Added missing Url.rewrite in response.middleware. Used a global variable to detect if functions is used in Navigationfallback.
Fixes #261


sgollapudi77:
Also Fixes #286

sgollapudi77:
@manekinekko is this test okay? 
If yes then we can close this PR, if not can you help me in adding some tests to verify the change.

manekinekko:
@sgollapudi77 I was referring to the cypress tests. But also, having unit tests is also highly recommended.

PR #294, opened 2021-09-10
feat: Add support for swa-cli.config.json file

### LINKED ISSUE #20: Add swa.json config file
As a user, I'd like to be able to start `swa` using a config file. All the CLI flags should be supported.

```json
{
  "options": {
   // CLI flags
  }
}
```

File name options:
- `swa.json`
- `swa.config.json`
- `.swarc`
- `swa` property in `package.json`

alexweininger:
### Feature:

This feature works as described by @anthonychu in https://github.com/Azure/static-web-apps-cli/issues/20#issuecomment-809577522.

* If only one configuration present in config file, running `swa start` will use this configuration.
* Otherwise you must use `swa start <config name>` to use a configuration. If `<config name>` is not found in the config file, no config will be used.
* Added the `--config` option, which allows the user to specify a specific path to a config file to use.
* Added the `--print-config` options, which prints the picked up configuration. Useful for user debugging.

I've updated the readme to document how to use a config file.

One thing to note: this feature can only either take options from the command line or the config file. It cannot merge or mix options. If a configuration from the config file is picked up and used no options passed in via command line will be respected.

This is a limitation of commander. More details at: https://github.com/tj/commander.js/issues/1584

I've also created a JSON Schema for the swa-cli.config.json file.

### Tests:
* Added unit tests
* Added e2e tests that function the same as the current e2e test, except use a config file rather than options passed in via terminal. Run these with `npm run e2e:config`

Closes #20

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -165,21 +165,84 @@ swa start http://localhost:3000 --swa-config-location ./my-app-source
 
 If you need to override the default values, provide the following options:
 
-| Options                          | Description                                             | Default   | Example                                              |
-| -------------------------------- | ------------------------------------------------------- | --------- | ---------------------------------------------------- |
-| `--app-location`                 | set location for the static app source code             | `./`      | `--app-location="./my-project"`                      |
-| `--app, --app-artifact-location` | set app artifact (dist) folder or dev server            | `./`      | `--app="./my-dist"` or `--app=http://localhost:4200` |
-| `--api, --api-artifact-location` | set the API folder or dev server                        |           | `--api="./api"` or `--api=http://localhost:8083`     |
-| `--swa-config-location`          | set the directory of the staticwebapp.config.json file. |           | `--swa-config-location=./my-project-folder`          |
-| `--api-port`                     | set the API server port                                 | `7071`    | `--api-port=8082`                                    |
-| `--host`                         | set the emulator host address                           | `0.0.0.0` | `--host=192.168.68.80`                               |
-| `--port`                         | set the emulator port value                             | `4280`    | `--port=8080`                                        |
-| `--ssl`                          | serving the app and API over HTTPS (default: false)     | `false`   | `--ssl` or `--ssl=true`                              |
-| `--ssl-cert`                     | SSL certificate to use for serving HTTPS                |           | `--ssl-cert="/home/user/ssl/example.crt"`            |
-| `--ssl-key`                      | SSL key to use for serving HTTPS                        |           | `--ssl-key="/home/user/ssl/example.key"`             |
-| `--run`                          | Run a command at startup                                |           | `--run="cd app & npm start"`                         |
-| `--devserver-timeout`            | The time to wait(in ms) for the dev server to start     | 30000     | `--devserver-timeout=60000`                          |
-| `--func-args`                    | Additional arguments to pass to `func start`            |           | `--func-args="--javascript"`                         |
+| Options                          | Description                                             | Default                 | Example                                              |
+| -------------------------------- | ------------------------------------------------------- | ----------------------- | ---------------------------------------------------- |
+| `--app-location`                 | set location for the static app source code             | `./`                    | `--app-location="./my-project"`                      |
+| `--app, --app-artifact-location` | set app artifact (dist) folder or dev server            | `./`                    | `--app="./my-dist"` or `--app=http://localhost:4200` |
+| `--api, --api-artifact-location` | set the API folder or dev server                        |                         | `--api="./api"` or `--api=http://localhost:8083`     |
+| `--swa-config-location`          | set the directory of the staticwebapp.config.json file. |                         | `--swa-config-location=./my-project-folder`          |
+| `--api-port`                     | set the API server port                                 | `7071`                  | `--api-port=8082`                                    |
+| `--host`                         | set the emulator host address                           | `0.0.0.0`               | `--host=192.168.68.80`                               |
+| `--port`                         | set the emulator port value                             | `4280`                  | `--port=8080`                                        |
+| `--ssl`                          | serving the app and API over HTTPS (default: false)     | `false`                 | `--ssl` or `--ssl=true`                              |
+| `--ssl-cert`                     | SSL certificate to use for serving HTTPS                |                         | `--ssl-cert="/home/user/ssl/example.crt"`            |
+| `--ssl-key`                      | SSL key to use for serving HTTPS                        |                         | `--ssl-key="/home/user/ssl/example.key"`             |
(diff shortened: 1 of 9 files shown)


readme.md
```diff
@@ -165,21 +165,84 @@ swa start http://localhost:3000 --swa-config-location ./my-app-source
...
+
+These configurations can be run with `swa start static` and `swa start devserver`.
+
+### Validation
+
+You can validate your `swa-cli.config.json` with a JSON Schema like so:
+
+```json
+{
+  "$schema": "https://raw.githubusercontent.com/Azure/static-web-apps-cli/main/schema/swa-cli.config.schema.json",
```
manekinekko, line 240: Suggestion/idea: host this config schema file somewhere on docs/ACOM?

cc @anthonychu @simonaco

src/core/utils/cli-config.ts
```diff
@@ -0,0 +1,53 @@
...
+export const swaCliConfigFilename = "swa-cli.config.json";
+
+export async function getFileOptions(context: string, configFilePath: string): Promise<SWACLIConfig & { context?: string }> {
+  if (!fs.existsSync(configFilePath)) {
+    return {};
+  }
+
+  const cliConfig = await tryParseSwaCliConfig(configFilePath);
+  if (cliConfig.configurations === undefined) {
+    logger.error(`${swaCliConfigFilename} is missing the "configurations" property. No options will be loaded.`);
```
manekinekko, line 16:
```suggestion
    logger.warn(`${swaCliConfigFilename} is missing the "configurations" property. No options will be loaded.`);
```

src/swa.d.ts
```diff
@@ -68,7 +68,7 @@ declare type SWACLIConfig = GithubActionWorkflow & {
   swaConfigFilename?: "staticwebapp.config.json";
   swaConfigFilenameLegacy?: "routes.json";
   app?: string;
-  api?: string;
+  apiLocation?: string;
```
manekinekko, line 71: any reason for this renaming?

schema/swa-cli.config.schema.json
```diff
@@ -0,0 +1,96 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "properties": {
+    "configurations": {
+      "additionalProperties": {
+        "allOf": [
+          {
+            "properties": {
+              "apiLocation": {
```
manekinekko, line 9: `apiLocation` and `app` should follow the same naming convention. This would make it less confusing for users.

Either (`apiLocation` and `appLocation`) or (`api` and `app`).

schema/swa-cli.config.schema.json
```diff
@@ -0,0 +1,96 @@
...
+              },
+              "apiPort": {
+                "description": "API backend port",
+                "type": "number"
+              },
+              "apiPrefix": {
+                "enum": ["api"],
+                "type": "string"
+              },
+              "app": {
```
manekinekko, line 21: same suggestion as above.

manekinekko (changes requested):

src/swa.d.ts
alexweininger: https://github.com/Azure/static-web-apps-cli/blob/main/src/cli/index.ts#L34

```ts
.option("--api, --api-location <apiLocation>", "set the API folder or Azure Functions emulator address", DEFAULT_CONFIG.apiLocation)
```

In the option we specify the option name as `apiLocation`. And this is what the property is named in `cli.opts()`. So in order to easily override this value, I've matched the property names.

We could add logic to allow us to use either `api` or `apiLocation` in the config file, but that might confuse users even more.

schema/swa-cli.config.schema.json
alexweininger: I think the issue comes from this:

`--api` and `--api-location` are the same option. Check my above comment for why I'd like to use `apiLocation`.

However, `--app` and `--app-location` are NOT the same option. As one might expect. Rather `--app` and `--app-artifact-location` are the same. And `--app-location` is a separate option from `--app`.

So we actually need three:
1. `app` or `appArtifactLocation`
2. `api` or `apiLocation`
3. `appLocation`

manekinekko (approved):

src/swa.d.ts
manekinekko: That makes sense. If you think this is gonna be confusing for users, then we can remove the shorthand flags `app` and `api` and just leave the flags that matched the ones SWA uses.

WDYT?

schema/swa-cli.config.schema.json
manekinekko: See https://github.com/Azure/static-web-apps-cli/pull/294#discussion_r711582369

schema/swa-cli.config.schema.json
```diff
@@ -0,0 +1,100 @@
...
+              },
+              "apiPort": {
+                "description": "API backend port",
+                "type": "number"
+              },
+              "apiPrefix": {
+                "enum": ["api"],
+                "type": "string"
+              },
+              "appArtifactLocation": {
```
manekinekko, line 21: Question: Is there an option to mark `appArtifactLocation` as deprecated? In SWA, this argument has been renamed to `outputLocation`.

alexweininger: If we mark it as deprecated would you want to include an option for `outputLocation`?

alexweininger: And the best way I found for marking a property as deprecated would be to just include it in the property description.

src/swa.d.ts
alexweininger: I think that would be more clear. Do you want to make that change in this PR or a new PR?

alexweininger: This would be a breaking change for any commands already setup to use `--api` and `--app`, are we ok with that?

manekinekko: I think this could be done in a different PR. 

FWIW, the CLI is still in _preview_ stage. But we can:
- plan any breaking change in version X+1.
- Notify the users (in version X) about this upcoming breaking change, if we detect any deprecated flag/command.

cc @Reshmi-Sriram

alexweininger: Sounds good. I will leave this PR as-is and plan on making these changes in another PR after this one is merged.

schema/swa-cli.config.schema.json
manekinekko: That'd be enough I think.

manekinekko (approved):
LGTM. cc @Reshmi-Sriram

sgollapudi77 (approved):

PR #321, opened 2021-09-29
move list of ports closer to the corresponding step

simonaco:


manekinekko (approved):

manekinekko:
Thank you @simonaco for this change. One last thing, it'd would be fantastic if you can edit your commit message and add `docs` prefix, like so:

```
docs: move list of ports closer to corresponding step
```

This would make our **Semantic Pull Request** check happy 😃

sgollapudi77 (approved):

Issue #353, opened 2021-11-25
Install azure-functions-core-tools when the user is trying to run swa start with the --api-location flag


simonaco:
**Is your feature request related to a problem? Please describe.**
Currently users have to install azure-functions-core-tools if they need to run the API.  

**Describe the solution you'd like**
When users are trying to run swa start --api-location, detect if they don't have azure-functions-core-tools installed and ask if they want to install it. This would lead to a better user experience and faster path to success.

**Describe alternatives you've considered**
Install the tools separately

**Additional context**
Installing the core tools separately requires context switching and many times users are not successful because they don't land on the right version or the simplest install command. First time users might not know why they need it and why so they might skip this step. 

recommended install using npm:

npm i -g azure-functions-core-tools@3 --unsafe-perm true

anthonychu:
Probably should install v4. Ideally it’ll select based on language and version to match the service but maybe just install v4 by default if func is not available, and user can downgrade if needed.

sinedied:
Is NPM still a valid/recommended way of installing core tools? I remember having issues with the core tools install silently failing (ie core tools not being available despite the package being installed) depending of the OS used.

The NPM package's readme itself states to use `brew` on MacOS or `dpkg` on Linux.

anthonychu:
Not sure if it’s worth investigating but theoretically SWA cli can download the func binary and run it, and it doesn’t need to depend on a separately installed version. This is more work, but there are advantages such as using the right version of core tools for the language in the project (.NET 6 requires v4, Node currently uses v3) and always keep the core tools up to date. Example of how to download the binary and start the right one: https://github.com/anthonychu/funcvm

sinedied:
Just giving a heads up as I'm currently working on direct download of the func tools correct version based on @anthonychu's suggestions 🙂

anthonychu:
Nice. Just a heads up that we might want to reach out to functions team to ensure we are looped in if there are any planned changes to how this works. I’m not aware of any plans to change the feed and download methods but should confirm.

sinedied:
I've sent the PR for this 🙂 : https://github.com/Azure/static-web-apps-cli/pull/391

I have two questions now:
1. Should we include a CLI flag to clean-up downloaded core tools?
2. Should we check for newer Core Tools version (than installed locally) and automatically update it?

sinedied:
Also @anthonychu do you know someone we could reach out to regarding the feed question?

manekinekko:
[...]

PR #356, opened 2021-12-01
feat: check for func binary and suggest core tools install (#353)

sinedied:
Fix proposal for #353 

If `func` binary is not found in $path, it will show this message:
image

It would be possible to interactively propose the user to run the command for him, but currently it's just a copy/paste away and allows users to use their favorite package manager if it's not NPM.

The docs link may also help in case the core tools install with NPM is not working for whatever reason, but also to give more context.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,26 +1,26 @@
 {
   "name": "@azure/static-web-apps-cli",
   "version": "0.8.3-alpha",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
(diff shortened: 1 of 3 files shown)


sinedied:
I've updated the PR so it can be merged.

After discussion with @manekinekko , I'll work on a follow-up PR that will automatically install the core tools using @anthonychu 's approach, and fall back to this message only when it fails.

manekinekko (approved):

PR #358, opened 2021-12-01
feat: add --open option (closes #357)

### LINKED ISSUE #357: [Suggestion] Automatically open browser after the swa server is ready
When using an API and external dev server (like for React, Angular, Vue...) the CLI shows a bunch of different URLs and users often open the wrong one in their browsers, ending up with the API and auth emulator not being available.

I would suggest adding a new flag `--open` that automatically opens the right URL (like `localhost:4280`) in their browser. This behavior is used by other similar dev server tools, like the `ng` CLI or `vue-cli`.

Making the correct URL more proeminent in the logs could also be an option, but when an API is also started it would end up being buried in API logs anyways, so the issue would remain.

sinedied:
Note that the default behavior remains unchanged.
Adding `--open` will automatically open the SWA server in the default browser.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "http-proxy": "^1.18.1",
         "internal-ip": "^6.2.0",
         "node-fetch": "^2.6.6",
+        "open": "^8.4.0",
         "ora": "^5.4.1",
         "rimraf": "^3.0.2",
         "serve-static": "^1.14.2",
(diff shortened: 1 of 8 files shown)


manekinekko (approved):

Reshmi-Sriram:
Hi @sinedied @manekinekko 
Thank you for adding this flag. Would you mind making the required changes in the [Readme](https://github.com/Azure/static-web-apps-cli#configuration) section as well so that people are aware of the additional option they have? TIA! :)

manekinekko:
Updated the readme in https://github.com/Azure/static-web-apps-cli/commit/feac33654dd3700d6a61707fd1dbece3e4aaf32e

PR #391, opened 2022-02-25
feat: install core tools automatically when needed (closes #353)

### LINKED ISSUE #353: Install azure-functions-core-tools when the user is trying to run swa start with the --api-location flag

sinedied:
Best description is a short demo:
func2

🎉 

What this PR does:
- Check if Core Tools are installed globally
- If not or if the global Core Tools version doesn't match the required Node.js runtime, it will install a local version shared among all CLI instances (installed in `~/.swa/core-tools`
- Multiple major versions of local Core Tools can coexist, for example v3 if you have a Node v12 project and v4 for a Node v16 projects (if you're using NVM for example)
- Use local Core Tools version when available


src/core/func-core-tools.ts
```diff
@@ -0,0 +1,217 @@
...
+import { promisify } from 'util';
+import fetch from 'node-fetch';
+import unzipper from 'unzipper';
+import cliProgress from 'cli-progress';
+import { logger } from "./utils/logger";
+
+const RELEASES_FEED_URL = 'https://functionscdn.azureedge.net/public/cli-feed-v4.json';
+const DEFAULT_FUNC_BINARY = 'func';
+const VERSION_FILE = '.release-version';
+const LOCAL_CORE_TOOLS_FOLDER = '.swa/core-tools';
```
manekinekko, line 15: Can we store the `func` binary in the node_modules (or similar)? I'd like us to avoid maintaining a new config folder.

Also, users may accidentally commit this binary to their git history.

src/core/func-core-tools.ts
```diff
@@ -0,0 +1,217 @@
...
+  version: string;
+  url: string;
+}
+
+interface CoreToolsZipInfo {
+  OS: string;
+  Architecture: string,
+  downloadLink: string;
+  size: string;
+}
```
manekinekko, line 28: Can we move these types to https://github.com/Azure/static-web-apps-cli/blob/main/src/swa.d.ts?

src/core/func-core-tools.ts
```diff
@@ -0,0 +1,217 @@
...
+  const release = await getLatestCoreToolsRelease(version);
+
+  // Make sure we start from a clean folder
+  if (fs.existsSync(dest)) {
+    removeDownloadedCoreTools(version);
+  } else {
+    fs.mkdirSync(dest, { recursive: true });
+  }
+
+  await downloadAndUnzipPackage(release.url, dest);
```
manekinekko, line 163: We need to make sure that the binary is not corrupted. Can we check the SHA2 sum against the one from the feed definition?

src/core/func-core-tools.ts
```diff
@@ -0,0 +1,217 @@
+import os from 'os';
```
manekinekko, line 1: We need additional unit tests for this new file.

manekinekko (changes requested):

src/core/func-core-tools.ts
sinedied: It's located in the user folder at `~/.swa/core-tools`, so it's shared among projects 😉 
Also no risk of commiting it by mistake there.

sinedied:
All comments should have been resolved, there're now more testing code that actual code 😅

PR #406, opened 2022-03-11
fix: ignore invalid workflow files

### LINKED ISSUE #363: SWA CLI requires GitHub Actions workflow file + specific jobs
Before filing this issue, please ensure you're using the latest CLI by running `swa --version` and comparing to the latest version on [npm](https://www.npmjs.com/package/@azure/static-web-apps-cli).

My SWA version is `0.8.2`.

**Are you accessing the CLI from the default port `:4280` ?**
- [ ] No, I am using a different port number (--port) and accessing the CLI from that port
- [x] Yes, I am accessing the CLI from port `:4280`

> Make sure you are accessing the URL printed in the console when running!

**Describe the bug**

* I've got the GitHub Actions workflow file for SWA. But it's refactored with the `workflow_call` event, so that both `build_and_deploy_job` and `close_pull_request_job` jobs are located in a separate workflow file.
* When I run the SWA CLI, it throws an error like:

```bash
C:\Users\justi\AppData\Roaming\nvm\v16.13.0\node_modules\@azure\static-web-apps-cli\dist\core\utils\workflow-config.js:63
        throw Error(`missing property 'jobs.build_and_deploy_job' in the SWA workflow file "${githubActionFile}". Make sure it's a valid SWA workflow file.`);
```

If I restore the job `build_and_deploy_job` to the original GitHub Actions workflow, it works fine.

**To Reproduce**
1. Change your GitHub Actions workflow filename for ASWA from `azure-static-web-apps-xxxx-xxxx-xxxx.yml` to something else (`hello-world.yml` for example), and SWA CLI will work fine.
2. Open your GitHub Actions workflow file for ASWA, whose filename starts with `azure-static-web-apps-`, change the job name from `build_and_deploy_job` to something else (`hello_world` for example), and SWA CLI will throw the error above.

**Expected behavior**

It should run without an error.

**Screenshots**

image

 - OS: Windows 11
 - Browser N/A
 - Version 0.8.2 (SWA-CLI)

**Additional context && Possible workaround**
If you want to avoid this error, you should duplicate the GitHub Actions workflow file for ASWA and update the event trigger with a "non-existing" directory path. Here's the trick:

* Duplicate original ASWA GHA workflow file, `azure-static-web-apps-xxxx-xxxx-xxxx.yml` to `azure-static-web-apps-swa-cli.yml`
* Leave your original ASWA GHA workflow file as it is, but update your dup'd workflow file like below:

```yaml
name: Azure Static Web Apps CI/CD

on:
  push:
    branches:
    - main
    paths:
    # Add this path which doesn't exist in your repository
    - "notfound/**"
  pull_request:
    types: [opened, synchronize, reopened, closed]
    branches:
    - main
    paths:
    # Add this path which doesn't exist in your repository
    - "notfound/**"

jobs:
  build_and_deploy_job:
  ...
```

Add "non-existing" path item like above (`notfound/**` for example).

Then, it tricks your SWA CLI and won't be triggered by either `push` or `pull_request` trigger. Here's the sample repository: https://github.com/devrel-kr/integration-villain/tree/main/.github/workflows

manekinekko:
Closes #363 

## Context:
We scan for GitHub Action workflow files (if present) in order to extract the build configuration. The way we validate the workflow file is based on the default templates that SWA generates when creating a new project. 

Initially, we threw an exception if the workflow file is not valid. With the new fix, we now simply show a warning and keep going.

### DIFF
--- a/cypress/fixtures/blazor-starter/swa-cli.config.json
+++ b/cypress/fixtures/blazor-starter/swa-cli.config.json
@@ -3,6 +3,7 @@
   "configurations": {
     "app": {
       "context": "http://localhost:5000",
+      "appLocation": "./cypress/fixtures/blazor-starter",
       "apiLocation": "./cypress/fixtures/blazor-starter/Api",
       "run": "dotnet run --project ./cypress/fixtures/blazor-starter/Client",
       "port": 1234,
--- a/cypress/fixtures/static/.github/workflows/azure-static-web-apps-NON-CONFORM.yml
+++ b/cypress/fixtures/static/.github/workflows/azure-static-web-apps-NON-CONFORM.yml
@@ -1,0 +1,27 @@
+name: Azure Static Web Apps CI/CD
+
+on:
+  push:
+    branches:
+    - main
+    paths:
+    - "wrapperapp/**"
+  pull_request:
+    types: [opened, synchronize, reopened, closed]
+    branches:
+    - main
+    paths:
+    - "wrapperapp/**"
+
+jobs:
+  call_aswa_flow:
+    uses: devrel-kr/integration-villain/.github/workflows/build-aswa.yaml@main
+    with:
+      event_name: ${{ github.event_name }}
+      event_action: ${{ github.event.action }}
+      app_location: "wrapperapp/Wrapper.WasmApp"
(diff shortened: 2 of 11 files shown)


fhinkel (approved):
Rubberstamp lgtm

PR #416, opened 2022-03-20
feat: swa deploy 🚀

manekinekko:
Deploy using a deployment token
```
  swa deploy --output-location ./app/dist/ --api-location ./api/ --deployment-token <token>
```

Deploy using a deployment token from env
```
  SWA_CLI_DEPLOYMENT_TOKEN=123 swa deploy --output-location ./app/dist/ --api-location ./api/
```

Deploy using swa-cli.config.json file
```
  swa deploy myconfig
  swa deploy
```

Run in dry-run mode
```
  swa deploy --dry-run
```

https://user-images.githubusercontent.com/1699357/159264654-ee0bc9aa-9af5-4e3c-bc33-ab18746f371e.mp4


src/cli/commands/deploy.ts
```diff
@@ -0,0 +1,196 @@
...
+          path.sep + DEFAULT_CONFIG.apiPrefix!
+        }" but the --api-location option was not provided. API will not be deployed!`,
+        "swa"
+      );
+    } else {
+      logger.warn(`No API found. Skipping...`, "swa");
+    }
+  } else {
+    logger.log(`Deploying API from folder:`, "swa");
+    logger.log(`  ${chalk.green(apiFolder)}`, "swa");
```
aaronpowell, line 44: Message here doesn't show the API folder that is being used, it'll always be the default one.

Maybe line 30 needs to be updated to have the `apiFolder` better reflecting the structure.

aaronpowell (changes requested):

src/cli/commands/deploy.ts
manekinekko: Fixed in the `ga` branch. Thank you @aaronpowell

PR #430, opened 2022-04-04
fix: improve query params logic if dev server

### LINKED ISSUE #429: When requesting a dynamic webpage from the dev site the proxy strips off the query parameters
- [X] Yes, I am accessing the CLI from port `:4280`

**Describe the bug**
When the browser requests a webpage that has query parameters from the backend/dev website (not the api endpoint or the auth redirects) the SWA http proxy removes the query parameters and therefore dynamic development servers cannot operator properly. 

I ran into this when using vite as the development server for a SFC driven VueJS 3 site.

**To Reproduce**
1. Follow the official VueJS Quick Start: https://vuejs.org/guide/quick-start.html to create an SFC 
2. Modify the package.json file to add the expected npm script `"serve": "vite --port 3000",`
3. Start the development server using the default SWA configuration
4. Note that the styles are not received by the browser and that the request for the inline styles return the default script not the style information.

** Details of my specific issue
I did some digging and in my project and isolated the following request:
`http://localhost:4280/src/components/Reports.vue?vue&type=style&index=0&lang.css)`
was being rewritten/proxied to the dev server as
`http://localhost:4280/src/components/Reports.vue`

**Expected behavior**
I expect that requests to the website endpoint server are passed through as is.

- OS: Windows 11
 - Browser: Chrome
 - Version:
   - Chrome: Version 100.0.4896.60 (Official Build) (64-bit)
   - SWA: 0.8.3

**Additional context**
I have worked around this issue by commenting out the following line https://github.com/Azure/static-web-apps-cli/blob/8f0045f58c6453f1ed89f900b900f5ac4ac61ac0/src/msha/middlewares/response.middleware.ts#L216 but It looks like this was introduced in the following PR:  https://github.com/Azure/static-web-apps-cli/pull/236 (I'm seeing src/msha/middlewares/response.middleware.ts line 100 (102 in 0.8.3) but have not managed to tease out the root cause)

manekinekko:
Closes #429


sinedied (approved):
LGTM

PR #432, opened 2022-04-06
chore(cli): update json schema (closes #422)

### LINKED ISSUE #422: Proposal: swa-cli.config.json format evolution to support new commands
With the upcoming changes on the SWA CLI and the new commands being developed, it is currently a problem that we cannot set different values for options flags depending of the running command.

Concrete example of the problem:
- For `swa start`, I would like to use a dev server (`swa start http://localhost:4200`) which means the `context` value of my configuration would be `http://localhost:4200`.
- For `swa deploy`, I need to specify the output folder (`swa deploy ./dist`) which means the `context` value of my configuration would need to be `./dist`.

**Problem:** currently we can only set 1 `context` property, and we have to way to differentiate its value depending of command used.

The proposed new format would be backward compatible with the current one, but would add an extra level to allow defining or overriding any option for a given command:

Screenshot 2022-03-28 at 16 21 38

With this proposition, options could still be defined "globally" (ie for all commands), and/or could be defined for a particular command. This would allow to handle use cases such as the example above, while not adding too much complexity and keeping compatibility with current format. Having the option of using both global and command-specific option also have the benefit of avoiding duplication for options that are identical between commands.

With @manekinekko, we've considered all these alternative approaches:
1. Adding a dedicated `overrides` property to allow overriding specific settings for each command => unneeded extra level, as we'll probably avoid having commands named the same as cli flags to avoid any confusion
1. Force an extra level for each command to have its own options => breaks compatibility, will need to duplication some options between commands
1. Add optional CLI flags that can be used as an alternative syntax in place of command context => this was the case with the start command and `--output-location` and cause confusion as it was possible to use both context and option, so it was removed
1. Add extra settings in the configuration that do no exists as CLI flags, but are injected in place of the context argument for each command (for example, outputLocation or startContext, deployContext) => having settings that do not match the CLI options would be confusing
1. Use existing `configurations` to create named profiles for each command, ie have a `app:start`, `app:deploy` etc configurations => would need reworking the configuration logic and make it prone to errors for grouping commands under the same "app" name
1. Allow to prefix configuration options with the name of the command to define or override it, for example `start:context` and `deploy:context` => would be hard to keep track of all settings related to given command

**Additional context**
With no pun intended, this may be also a good chance to rethink the usage of the `context` option in the configuration file.
The `context` option is currently exposed as an option in the configuration file (to use like `swa start <context>` for example), but it was initially meant to be an implementation detail to my understanding. So maybe, we could also reconsider option 3. in addition to the current proposal, to remove the need to expose the `context` setting in the configuration file.

This would allow to be more consistent with SWA options in the GitHub actions and the portal, as `context` have no meaning there but for example `output-location` has.
To avoid the initial confusion issue that resulted in its removal, we could simply detect if both context and its alternative options are set at the same time when calling the CLI and throw and error. 
For example, running `swa start http://localhost:4200 --output-location ./dist` would cause an error as both the context (`http://localhost:4200`) and `--output-location` flag are defined at the same time.

sinedied:
Cleaned unused settings and constants from schema:
- appArtifactLocation
- build (superseded by appBuildCommand and apiBuildCommand)
- apiPrefix
- swaConfigFilename
- swaConfigFilenameLegacy
- customUrlScheme
- overridableErrorCode

We can't enforce proper schema exclusion for extra properties as we need to migrate to draft/2019-09 to be able to use `unevaluatedProperties` for that. But draft/2019-09 version isn't currently properly supported by VS Code, so it's not the good time to migrate.

I specifically enforced exclusion of `deploymentToken` from config file though.

### DIFF
--- a/schema/swa-cli.config.schema.json
+++ b/schema/swa-cli.config.schema.json
@@ -8,13 +8,13 @@
             "$ref": "#/$defs/globalParameters"
           },
           {
-            "$ref": "#/$defs/workflowParameters"
-          },
-          {
             "$ref": "#/$defs/startCommandParameters"
           },
           {
             "$ref": "#/$defs/loginCommandParameters"
+          },
+          {
+            "$ref": "#/$defs/loginSharedCommandParameters"
           },
           {
             "$ref": "#/$defs/initCommandParameters"
@@ -36,6 +36,9 @@
                     "$ref": "#/$defs/loginCommandParameters"
                   },
                   {
+                    "$ref": "#/$defs/loginSharedCommandParameters"
+                  },
+                  {
                     "$ref": "#/$defs/contextParameters"
                   }
                 ]
@@ -48,9 +51,6 @@
                 "allOf": [
                   {
                     "$ref": "#/$defs/globalParameters"
-                  },
-                  {
-                    "$ref": "#/$defs/workflowParameters"
                   },
                   {
                     "$ref": "#/$defs/startCommandParameters"
@@ -87,7 +87,7 @@
                     "$ref": "#/$defs/globalParameters"
                   },
                   {
-                    "$ref": "#/$defs/workflowParameters"
+                    "$ref": "#/$defs/loginSharedCommandParameters"
                   },
                   {
                     "$ref": "#/$defs/deployCommandParameters"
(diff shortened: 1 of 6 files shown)


sinedied:
@alexweininger can you check if you see anything potentially breaking for the VS Code extension?

alexweininger:
> @alexweininger can you check if you see anything potentially breaking for the VS Code extension?

Will do, thanks for notifying me.

manekinekko (approved):

sinedied:
This PR will closes https://github.com/Azure/static-web-apps-cli/issues/422 BTW.

PR #435, opened 2022-04-07
feat: add custom persistence cache options

manekinekko:


scripts/copy-assets.js
```diff
@@ -18,6 +18,8 @@ try {
     path.join("src", "public", "401.html"),
     path.join("src", "public", "404.html"),
     path.join("src", "public", "403.html"),
+    path.join("src", "public", "codeFlowResult", "index.html"),
```
manekinekko, line 21: this is not needed anymore. I am gonna remove it later today.

PR #440, opened 2022-04-10
fix: set a random port if --port is already taken

manekinekko:
image

cc @Reshmi-Sriram

### DIFF
--- a/src/core/utils/net.ts
+++ b/src/core/utils/net.ts
@@ -1,9 +1,8 @@
 import chalk from "chalk";
+import getPort from "get-port";
 import net from "net";
 import ora from "ora";
 import waitOn from "wait-on";
-import getPort from "get-port";
-import { DEFAULT_CONFIG } from "../../config";
 import { confirmChooseRandomPort } from "../prompts";
 import { logger } from "./logger";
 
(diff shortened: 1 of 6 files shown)


src/core/utils/net.ts
```diff
@@ -2,24 +2,51 @@ import chalk from "chalk";
...
+  host = DEFAULT_CONFIG.host,
+  port = DEFAULT_CONFIG.port,
+}: {
+  host?: string;
+  port?: number;
+}): Promise<number> {
+  port = Number(port) as number;
+  logger.silly(`Checking if ${host}:${port} is accepting TCP connections...`);
+
+  // Check if current port is beyonf the MAX valid port range
```
sinedied, line 28:
```suggestion
  // Check if current port is beyond the MAX valid port range
```

src/core/utils/net.ts
```diff
@@ -2,24 +2,51 @@ import chalk from "chalk";
...
 /**
  * Check if a given remote address and port are accepting TCP connections.
  * @param Object host and port of the server to check.
- * @returns True if the given server is accepting TCP connections. False otherwise.
+ * @returns The resolved port number if the given server is accepting TCP connections. 0 if the port is already taken.
  */
-export function isAcceptingTcpConnections({ host = "127.0.0.1", port = 80 }: { host?: string; port?: number }) {
-  return new Promise<boolean>((resolve) => {
-    const socket = net.createConnection(port, host);
+export function isAcceptingTcpConnections({
```
sinedied, line 18:
```suggestion
export async function isAcceptingTcpConnections({
```

src/core/utils/net.ts
```diff
@@ -2,24 +2,51 @@ import chalk from "chalk";
...
+  host = DEFAULT_CONFIG.host,
```
sinedied, line 19: host/port should come from the commander default, so there should be no need to have a fallback wired to DEFAULT_CONFIG here again?

src/core/utils/net.ts
```diff
@@ -2,24 +2,51 @@ import chalk from "chalk";
...
+        resolve(port as number);
       })
-      .once("connect", () => {
-        resolve(true);
+      .once("connect", async () => {
+        logger.warn(`Port ${port} is already taken!`);
+        const portNumber = await confirmChooseRandomPort();
+
+        if (portNumber) {
+          resolve(await getPort());
```
sinedied, line 48: No need for await, you can resolve a promise directly
```suggestion
          resolve(getPort());
```

sinedied (approved):
Nothing blocking, but I made a few suggestions where things could be improved

PR #441, opened 2022-04-11
feat: support for self-generating ssl certificates

### LINKED ISSUE #213: Ease the use of self-signed certs
SSL is required when using the media functions, for example to use Azure Communication Services. It is possible to create a local cert, however this is not easy to maintain when multiple contributors work on the same project. You don't want to check a local self-signed cert in your version control.

Dotnet CLI has [a function to generate self-signed certificates](https://docs.microsoft.com/en-us/dotnet/core/additional-tools/self-signed-certificates-guide) and it would be great if there was similar functionality in static-web-apps-cli. Or at least, if it was possible to plug in a tool to generate (and trust) these certs on the fly.

See https://www.hanselman.com/blog/developing-locally-with-aspnet-core-under-https-ssl-and-selfsigned-certs.

A better solution, or improved documentation how to handle this issue

Just use HTTP and use ngrok or GitHub Codespaces as a SSL reverse proxy during development.

manekinekko:
Closes #213


src/cli/commands/start.ts
```diff
@@ -212,8 +213,22 @@ export async function start(startContext: string | undefined, options: SWACLICon
   }
 
   if (options.ssl) {
-    if (options.sslCert === undefined || options.sslKey === undefined) {
-      logger.error(`SSL Key and SSL Cert are required when using HTTPS`, true);
+    if (options.sslCert === undefined && options.sslKey === undefined) {
+      logger.warn(`WARNING: Using built-in UNSIGNED certificate. DO NOT USE IN PRODUCTION!`);
```
sinedied, line 217: nit: not sure the production mention is needed here, as it's for `swa start` which is just for local emulation ;)

src/cli/commands/start.ts
```diff
@@ -212,8 +213,22 @@ export async function start(startContext: string | undefined, options: SWACLICon
...
+      const pemFilepath = await getCertificate({
+        selfSigned: true,
+        days: 365,
+        commonName: DEFAULT_CONFIG.host,
```
sinedied, line 221: should use options.host here?
```suggestion
        commonName: options.host,
```

src/core/ssl.ts
```diff
@@ -0,0 +1,136 @@
...
+    const pemContent = [unsignedCertificate.csr, unsignedCertificate.serviceKey, unsignedCertificate.certificate].join("\n");
+
+    await mkdir(cacheDir, { recursive: true });
+    await writeFile(cachePath, pemContent);
+
+    return cachePath;
+  }
+}
+
+function createCertificate(options: PEMOptions): Promise<SSLCertificate> {
```
sinedied, line 126: Functions returning promises should be async
```suggestion
function async createCertificate(options: PEMOptions): Promise<SSLCertificate> {
```

sinedied (changes requested):
LGTM except one small issue if user specify custom host with --host

PR #457, opened 2022-04-29
feat: make positional arguments explicit

sinedied:
BREAKING CHANGES:
    
--api-location options cannot take an API server URL anymore, --api-server-url must now be used for this use case.


src/cli/commands/start.ts
```diff
@@ -51,15 +56,31 @@ export default function registerCommand(program: Command) {
...
-      console.warn(chalk.yellow("* cloud environment exactly. Always deploy and test your app in Azure. *"));
-      console.warn(chalk.yellow("************************************************************************"));
+    .action(async (positionalArg: string = `.${path.sep}`, _options: SWACLIConfig, command: Command) => {
+      const options = await configureOptions(positionalArg, command.optsWithGlobals(), command, "start");
+      if (!matchLoadedConfigName(positionalArg)) {
+        // If it's not the config name, it's either output location or dev server url
+        const isUrl = isHttpUrl(positionalArg);
+        if (isUrl) {
+          if (isUserOption('devServerUrl')) {
+            logger.error(`devServerUrl was set on both positional argument and option.`, true);
```
manekinekko, line 66: TODO: make this message more explicit.

src/cli/index.ts
```diff
@@ -42,6 +42,7 @@ export async function run(argv?: string[]) {
         .default(DEFAULT_CONFIG.verbose)
     )
     .option("--config <path>", "path to swa-cli.config.json file to use", path.relative(process.cwd(), swaCliConfigFilename))
+    .option("--configName <name>", "name of the configuration to use", undefined)
```
manekinekko, line 45:
```suggestion
    .option("--config-name <name>", "name of the configuration to use", undefined)
```

src/cli/commands/start.ts
```diff
@@ -179,6 +201,8 @@ export async function start(outputLocationOtHttpUrl: string, options: SWACLIConf
   // use any specific workflow config that the user might provide undef ".github/workflows/"
   // Note: CLI args will take precedence over workflow config
   try {
+    // TODO: not sure if we should still do this here, as config/user options should override
+    // over any options in the workflow config, but it seems to do the opposite here.
```
manekinekko, line 205: `readWorkflowFile` does a mixing of the user's options (CLI flags) and the local GitHub action (if available). This was done prior to the info of the user config. Feel free to refactor this section if needed.

src/core/utils/cli-config.ts
```diff
@@ -64,14 +75,15 @@ export async function getConfigFileOptions(
...
   // TODO: comparison here should be a constant, used also as the default context
   // for all commands taking a context argument
-  if (contextOrConfigEntry === `.${path.sep}`) {
+  if (configName === `.${path.sep}`) {
     const hasMultipleConfig = Object.entries(cliConfig.configurations).length > 1;
     if (hasMultipleConfig) {
       // Show as a log not warning because the user may want to use the default config
       logger.log(`Multiple configurations found in "${swaCliConfigFilename}", but none was specified.`);
-      logger.log(`Specify which configuration to use with "swa <command> <configurationName>"\n`);
+      logger.log(`Specify which configuration to use with "swa <command> --config-name <configName>"\n`);
```
manekinekko, line 86: Suggestion: can we list all found config names so the user can choose from?

manekinekko (changes requested):

src/cli/commands/start.ts
sinedied: Once everything else is fixed, I'll try loading the workflow config before everything else set by user (cli args, options...) in option.ts, this should be the most logical option and would also benefit deploy (if nothing else is set).

src/core/utils/cli-config.ts
sinedied: I think it's best from a DX perspective that we fall back to a default value instead of going interactive every time, as CLI user I would find it annoying. What would be even better though would be a way to set which configuration to use as default, that should be easy to do but more docs/tests/schema to update, I'll see if it can fits before GA

PR #463, opened 2022-05-02
feat: add command paths when needed

sinedied:


src/core/frameworks/detect.ts
```diff
@@ -52,6 +52,24 @@ export async function generateConfiguration(app?: DetectedFolder, api?: Detected
...
+    commandPath = hasSpaces(commandPath) ? `"${commandPath}"` : commandPath;
+    config.appBuildCommand = `cd ${commandPath} && ${config.appBuildCommand}`;
+  }
+
+  const apiRootPath = api && removeTrailingPathSep(api.rootPath);
+  if (apiRootPath && config.apiBuildCommand && apiRootPath !== config.appLocation) {
+    // If the final api location is not the same as the detected root path of the app,
+    // we need to adjust the build command to run in the correct path.
+    let commandPath = path.relative(config.appLocation!, apiRootPath);
+    commandPath = hasSpaces(commandPath) ? `"${commandPath}"` : commandPath;
```
manekinekko, line 69: why not just use `"${commandPath}"` by default? This would escape the path with or without spaces.

manekinekko (approved):

src/core/frameworks/detect.ts
sinedied: This makes the config the config less easier to read because of quotes escaping in the final JSON, so unless it's necessary I thought that it would be better to keep it cleaner.

PR #534, opened 2022-07-06
feat(start): allow insecure https connections to remote api hosts

danwatford:
Allow insecure connections to remote api hosts to handle cases where remote TLS host presents a wildcard certificate.

Change the host header in proxied requests from localhost to match the remote api host.

Relates to #523

### DIFF
--- a/src/cli/commands/start.ts
+++ b/src/cli/commands/start.ts
@@ -35,8 +35,21 @@ export default function registerCommand(program: Command) {
     .option("-a, --app-location <path>", "the folder containing the source code of the front-end application", DEFAULT_CONFIG.appLocation)
     .option("-i, --api-location <path>", "the folder containing the source code of the API application", DEFAULT_CONFIG.apiLocation)
     .option("-O, --output-location <path>", "the folder containing the built source of the front-end application", DEFAULT_CONFIG.outputLocation)
-    .option("-D, --app-devserver-url <url>", "connect to the app dev server at this URL instead of using output location", DEFAULT_CONFIG.appDevserverUrl)
-    .option("-is, --api-devserver-url <url>", "connect to the api server at this URL instead of using output location", DEFAULT_CONFIG.apiDevserverUrl)
+    .option(
+      "-D, --app-devserver-url <url>",
+      "connect to the app dev server at this URL instead of using output location",
+      DEFAULT_CONFIG.appDevserverUrl
+    )
+    .option(
+      "-is, --api-devserver-url <url>",
+      "connect to the api server at this URL instead of using output location",
+      DEFAULT_CONFIG.apiDevserverUrl
+    )
+    .option(
+      "-ik, --api-devserver-insecure",
+      "allow insecure connections to the API server. Useful when HTTPS API server uses wildcard certificates.",
+      DEFAULT_CONFIG.apiDevserverInsecure
+    )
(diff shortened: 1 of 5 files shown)


sinedied:
Hey @danwatford , thanks for your contribution!

I was thinking that maybe this could be something that could be always activated as it's a proxy to a dev server anyways? @manekinekko what do you think?

manekinekko:
> Hey @danwatford , thanks for your contribution!
> 
> I was thinking that maybe this could be something that could be always activated as it's a proxy to a dev server anyways? @manekinekko what do you think?

We are going to be adding support for `/api` proxying to [non-Azure Function APIs](https://github.com/Azure/static-web-apps-cli/issues/567) as well (AS, ACA, AKS...). I am not sure how this option will affect these remote connections, so I guess we can still provide this flag to users, but have it be `false` by default.

@danwatford have you tried using the `--ssl` option to enable SSL for local dev? Or this is not related?

danwatford:
> 
> @danwatford have you tried using the `--ssl` option to enable SSL for local dev? Or this is not related?

Hi @manekinekko 
The `--ssl` option only affects serving of the SWA front-end. To enable the CLI to proxy to an API hosted in a GitHub codespaces debug session or in Azure Functions, we need to change the behaviour of the http-proxy node module.

Issue #534 mentions how http-proxy doesn't handle wildcard certificates from remote API hosts. This PR is a workaround in that http-proxy is (optionally) instructed to ignore any remote host certificate authentication.

Assuming the new support for `/api` proxying to other APIs hosted remotely will continue to use http-proxy, then I think you will run in to the same problem with wildcard certificate handling and may still want to consider applying this PR.

However if there are plans to replace http-proxy with something that can handle wildcard certificates then this PR should not be necessary.

danwatford:
> Hey @danwatford , thanks for your contribution!
> 
> I was thinking that maybe this could be something that could be always activated as it's a proxy to a dev server anyways? 

Hi @sinedied ,

It might be acceptable to leave the insecure connection option enabled by default, but if we did that then we should probably log a warning to the user that we have not authenticated the connection to the remote API server.

I don't know if we can assume users will always be proxying to a development server. My preference would be to ensure a secure connection by default, but log some guidance about using the insecure connection option if they have any connectivity issues.

danwatford:
Some good(ish) news. While updating the documentation I wanted to capture the error text that users might see for problems connecting to remote API hosts, but couldn't recreate the problem that was intended to be solved by the --api-devserver-insecure command line option.

Testing against APIs accessed GitHub Codespaces public ports or via locally developed APIs accessed via Ngrok were no longer failing.

I tracked down the cause of the unintended successful connection to the `changeOrigin: true` option passed to http-proxy.

I think I had misinterpreted errors such as:
```
Hostname/IP does not match certificate's altnames: Host: localhost. is not in the cert's altnames: DNS:online.visualstudio.com, DNS:*.online.visualstudio.com, DNS:*.app.online.visualstudio.com, DNS:*.tunnels.vsengsaas.visualstudio.com
```
as http-proxy having issues with wildcard certificates. I had failed to notice the `Host: localhost` portion of the error message.

Conclusion: This PR can be closed and I will raise a new PR to apply the `changeOrigin: true` property for http-proxy and another PR to update the documentation on how to connect to a remote API server.

PR #564, opened 2022-09-16
fix(api): Node18 support

aaronpowell:
With the recent release of support for Node 18 in the Functions v4 runtime, we should update the tools check code to allow that.

### DIFF
--- a/src/core/func-core-tools.spec.ts
+++ b/src/core/func-core-tools.spec.ts
@@ -66,9 +66,12 @@ describe("funcCoreTools", () => {
       expect(isCoreToolsVersionCompatible(4, 16)).toBe(true);
       expect(isCoreToolsVersionCompatible(3, 16)).toBe(false);
       expect(isCoreToolsVersionCompatible(2, 16)).toBe(false);
-      expect(isCoreToolsVersionCompatible(4, 17)).toBe(false);
+      expect(isCoreToolsVersionCompatible(4, 17)).toBe(true);
       expect(isCoreToolsVersionCompatible(3, 17)).toBe(false);
       expect(isCoreToolsVersionCompatible(2, 17)).toBe(false);
+      expect(isCoreToolsVersionCompatible(4, 18)).toBe(true);
+      expect(isCoreToolsVersionCompatible(3, 18)).toBe(false);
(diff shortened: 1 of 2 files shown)


src/core/func-core-tools.spec.ts
```diff
@@ -66,9 +66,12 @@ describe("funcCoreTools", () => {
       expect(isCoreToolsVersionCompatible(4, 16)).toBe(true);
       expect(isCoreToolsVersionCompatible(3, 16)).toBe(false);
       expect(isCoreToolsVersionCompatible(2, 16)).toBe(false);
-      expect(isCoreToolsVersionCompatible(4, 17)).toBe(false);
+      expect(isCoreToolsVersionCompatible(4, 17)).toBe(true);
```
manekinekko, line 69: Can you share a link mentioning support for Node 17?

From the docs, it's only mentioning LTS versions:
image

aaronpowell: It isn't, but I was copying the pattern above where v4 would allow node15 (line 64) and v3 would allow node13 (line 60).

sgollapudi77 (approved):

sinedied (approved):

sulabh-msft (approved):

manekinekko (approved):

xxnickles:
Is this really supported? Just tried on Windows 11 (nodejs lts 18.12.1) and it cannot connect to the API endpoint (I am using the bring your functions approach with a function 4 project) However, swa works just find in node 16 (I had to downgrade my install). Am I missing anything here?

aaronpowell:
Assuming you're using the 1.0.3 release of SWA CLI it should be there

xxnickles:
> Assuming you're using the 1.0.3 release of SWA CLI it should be there

 I am using 1.0.3. Even though there are no explicit errors related to the nodejs version as there were in previous versions, the proxy is just not able to connect to the functions project when using node 18 lts ( ✖ Could not connect to "http://localhost:7071". Is the server up and running?) but works with node 16. I tried several times upgrading/downgrading and I have got the same results (at least in windows 11, I need to check it on fedora) Are you aware of any specific conditions (like having to use preview sdk for functions) to work with swa cli and node js 18? Just for additional piece information, I am using Blazor + dotnet functions

PR #578, opened 2022-10-03
feat(start): set host header in proxied requests to match remote API hosts

danwatford:
Changes the host header from localhost to the remote API host, permitting a host match to be found against the remote host's certificate.

Relates to #523

### DIFF
--- a/src/msha/handlers/function.handler.ts
+++ b/src/msha/handlers/function.handler.ts
@@ -70,6 +70,8 @@ export function handleFunctionRequest(req: http.IncomingMessage, res: http.Serve
     res,
     {
       target,
+      // Set the host header to match the remote API host.
+      changeOrigin: true,
     },
     onConnectionLost(req, res, target, "↳")
   );


sgollapudi77 (approved):

src/msha/handlers/function.handler.ts
```diff
@@ -70,6 +70,8 @@ export function handleFunctionRequest(req: http.IncomingMessage, res: http.Serve
     res,
     {
       target,
+      // Set the host header to match the remote API host.
+      changeOrigin: true,
```
manekinekko, line 74: Have you tried the local auth emulation with this option? I just want to make sure this option doesn't affect the cookies that are created locally. I don't think we set the domain in the cookie but just in case.

danwatford: No, I've not tested that case. If no one else if able to, I'll do so, but it will be a few days before I get a chance to work on swa-cli again.

danwatford: This has now been tested with the SWA CLI's authentication emulation. Test approach is described here - https://watfordconsulting.com/2022/10/05/access-the-azure-static-web-apps-clientprincipal/

Tested by running the auth-echo Azure Function both locally and remotely in a GitHub Codespace.

davide-bergamini-sevenit:
Please approve this, it's blocking for many. Thanks you!

jmlane:
@sulabh-msft This seems to be blocked waiting for your approval.

PR #579, opened 2022-10-03
docs: add guidance on connecting to a remote API host

danwatford:
Depends on #578


docs/www/docs/cli/swa-start.md
```diff
@@ -108,7 +118,7 @@ Here are the options you can use with `swa start`:
 - `-i, --api-location <path>`: the folder containing the source code of the API application
 - `-O, --output-location <path>`: the folder containing the built source of the front-end application. The path is relative to `--app-location` (default: ".")
 - `-D, --app-devserver-url <url>`: connect to the app dev server at this URL instead of using output location
-- `-is, --api-devserver-url <url>`: connect to the api server at this URL instead of using output location
+- `-is, --api-devserver-url <url>`: connect to the api server at this URL instead of using output location. Set `--api-location` to `remote` if using this option.
```
manekinekko, line 121: can't we set this automatically by the cli, or ignore it if the user is connecting to a remote address?

danwatford: Actually after testing some other swa-cli bits I realised that specifying an api-devserver-url option is not necessary.  I think the bit that confused me was following in cli/commands/start.ts which suggested api-location would be ignored if it didn't resolve to file location, but it was still necessary to specify api-location in order to get apiDevServerUrl checked.

https://github.com/Azure/static-web-apps-cli/blob/19c619ba4630ec809f2ddfdb6796afb6773ff337/src/cli/commands/start.ts#L186-L201

I now realise that even though a line about the api-location being missing was logged, the api-location was still passed to msha/server so was still accepted. 

I'll clean up the example and `api-devserver-url` option documentation.

What usecases is the `api-devserver-url` option needed for?

PR #626, opened 2023-01-09
fix: empty spaces uncorrectly detected as positional argument

sinedied:
Empty spaces are currently misdetected as positional arguments, for example `swa deploy     --output-location dist` will incorrectly interpret the space as a positional argument and print an error because it conflicts with the `output-location` option.

This PR fixes that.

### DIFF
--- a/src/cli/commands/build.ts
+++ b/src/cli/commands/build.ts
@@ -28,6 +28,7 @@ export default function registerCommand(program: Command) {
     .option("-I, --api-build-command <command>", "the command used to build your api", DEFAULT_CONFIG.apiBuildCommand)
     .option("--auto", "automatically detect how to build your app and api", false)
     .action(async (positionalArg: string | undefined, _options: SWACLIConfig, command: Command) => {
+      positionalArg = positionalArg?.trim();
       const options = await configureOptions(positionalArg, command.optsWithGlobals(), command, "build");
       if (positionalArg && !matchLoadedConfigName(positionalArg)) {
         if (isUserOption("appLocation")) {
--- a/src/cli/index.spec.ts
+++ b/src/cli/index.spec.ts
@@ -3,8 +3,8 @@ import { UpdateNotifier } from "update-notifier";
 import mockFs from "mock-fs";
 import { run } from "./index";
 
-jest.mock("child_process", () => ({
-  execSync: jest.fn(),
+jest.mock("./commands/build/build", () => ({
+  build: jest.fn(),
 }));
 
 const pkg = require("../../package.json");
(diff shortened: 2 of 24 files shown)


src/cli/commands/build.ts
```diff
@@ -28,6 +28,7 @@ export default function registerCommand(program: Command) {
     .option("-I, --api-build-command <command>", "the command used to build your api", DEFAULT_CONFIG.apiBuildCommand)
     .option("--auto", "automatically detect how to build your app and api", false)
     .action(async (positionalArg: string | undefined, _options: SWACLIConfig, command: Command) => {
+      positionalArg = positionalArg?.trim();
```
sulabh-msft, line 31: @sinedied Thanks! for raising the fix here. 

Will it be possible to add few tests for these changes ?

sinedied: Sure!

src/cli/index.spec.ts
```diff
@@ -53,4 +67,18 @@ describe("cli", () => {
...
+    const execSyncMock = jest.requireMock("child_process").execSync;
+    mockFs();
+    await run(["node", "swa", "build", "   app  ", "--app-build-command", "npm run something"]);
+    expect(execSyncMock.mock.calls[0][1].cwd).toBe("app");
+  });
+
+  it("should not interpret empty spaces as a positional argument", async () => {
+    const execSyncMock = jest.requireMock("child_process").execSync;
+    mockFs();
+    await run(["node", "swa", "build", "    ", "--app-build-command", "npm run something", "   "]);
```
sulabh-msft, line 81: Great thanks! @sinedied  for adding the test for `swa build`.. 

Just to confirm if these tests cover the changes we did in deploy.ts/start.ts if not will it make sense for us to add separate tests for those scenarios ?

sulabh-msft:
@sgollapudi77 / @rupareddy5-21  Can you also please review the PR ?

sulabh-msft (approved):
Approving the PR .. Please check if the tests added cover the changes for `deply` and `start`  as well.

src/cli/index.spec.ts
sinedied: The added logic is the same for build/start/deploy, so it should validate these changes, but not cover it in case of refactor of start/deploy.

Currently there's no tests (at all) for the cli commands deploy/start/login, hence why I reused the mocks made for the build command tests. Adding minimal tests for these commands would be useful, though not the scope of this PR.

sinedied:
Hmmm, seems in some Node environments jest fails to restore properly the "child_process" mocks. making the tests fail. I had a hard time finding a workaround already locally, I'll try to use another approach then and refactor the code a bit.

sinedied:
Tests are passing now.  I refactored the cli commands architecture a bit to split the registering (=doc and options parsing) and the execution parts. Testing individual commands and their components should be easier now.

sulabh-msft (approved):

sulabh-msft:
@sinedied looks like there is one merge conflict. Can you please resolve the merge conflict ?

sinedied:
@sulabh-msft I've updated the branch, no more conflict :)

Issue #610, opened 2022-11-23
SWA CLI stopping without error msg when using Blazor WASM dev server over https


StacyCash:
**Are you accessing the CLI from the default port `:4280` ?**

- [ ] No, I am using a different port number (`--port`) and accessing the CLI from that port
- [X] Yes, I am accessing the CLI from port `:4280`

**Describe the bug**
When trying to use the SWA CLI with a dotnet Blazor application running on https the SWA CLI quits when the index page is accessed (see below for traces from various node versions)

When accessing the http endpoint of the Blazor development server it works as expected

I've tried this on both Windows and macOS with the same result

**To Reproduce**

1. Clone the `https://github.com/Apress/beginning-azure-static-web-apps` repo
2. Open the `chapter-5` folder
3. Run `swa start https://localhost:5000 --api-location ./Api --run "dotnet watch run --project ./Client"`
4. Open `http://localhost:4280/`
5. SWA CLI quits without error (in the most recent versions of Node)

**Expected behavior**
The application should load in the browser as it does with the http endpoint

**Desktop (please complete the following information):**

- OS: Windows 11, macOS Ventura 13.0
- SWA CLI Version: 1.0.3

**Additional context**
When downgrading the SWA CLI to 1.0.2 the application works as expected

I tried downgrading Node to see if that had an impact. It did - but only in that running Node 14.x there is an error message displayed. If I build and run the SWA CLI on my machine removing the code that gracefully shuts down the SWA CLI then I can also get this error message on Node 16 (I didn't try on 18.x)

Downgrading Node to 14.21.1 I get the following:

``` ps
(node:12156) UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_PROTOCOL]: Protocol "https:" not supported. Expected "http:"
[swa]     at new NodeError (internal/errors.js:322:7)
[swa]     at new ClientRequest (_http_client.js:159:11)
[swa]     at Object.request (https.js:370:10)
[swa]     at Array.stream (C:\code\github\static-web-apps-cli\node_modules\http-proxy\lib\http-proxy\passes\web-incoming.js:126:74)
[swa]     at ProxyServer.<anonymous> (C:\code\github\static-web-apps-cli\node_modules\http-proxy\lib\http-proxy\index.js:81:21)
[swa]     at serveStaticOrProxyResponse (C:\code\github\static-web-apps-cli\dist\msha\middlewares\request.middleware.js:97:18)
[...]

PR #628, opened 2023-01-09
fix: providing separate agents for HTTP and HTTPS proxy servers

rupareddy5-21:
creating separate http and https agents depending on whether:
The user is hitting an http:// or https:// dev server.
The user is starting the CLI using --ssl or --no-ssl (by default)
Also, warning the user if:
They started the CLI using  --ssl but they are connecting to http:// dev server
They started the CLI with no SSL but they are connecting to https:// dev server

### DIFF
--- a/docs/www/docs/cli/swa-start.md
+++ b/docs/www/docs/cli/swa-start.md
@@ -160,6 +160,10 @@ Connect both front-end and the API to running development server
 swa start http://localhost:3000 --api-devserver-url http://localhost:7071
 ```
 
+:::info Note
+To make a secure connection, hit a https dev server only if you start the CLI using the flag --ssl.
+:::
+
 ## See Also
 
 - [swa](./swa)
--- a/src/cli/commands/start.ts
+++ b/src/cli/commands/start.ts
@@ -116,7 +116,6 @@ swa start http://localhost:3000 --api-devserver-url http://localhost:7071
     );
 }
 
-export var isSSL: boolean = false;
 export async function start(options: SWACLIConfig) {
   // WARNING:
   // environment variables are populated using values provided by the user to the CLI.
@@ -275,7 +274,6 @@ export async function start(options: SWACLIConfig) {
   }
 
   if (ssl) {
-    isSSL = ssl;
     if (sslCert === undefined && sslKey === undefined) {
       logger.warn(`WARNING: Using built-in UNSIGNED certificate. DO NOT USE IN PRODUCTION!`);
       const pemFilepath = await getCertificate({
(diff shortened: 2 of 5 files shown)


src/cli/commands/start.ts
```diff
@@ -112,6 +116,7 @@ swa start http://localhost:3000 --api-devserver-url http://localhost:7071
     );
 }
 
+export var isSSL: boolean = false;
```
manekinekko, line 119: commands should not export variables. Use `swaCLIEnv()` for that.

src/msha/server.ts
```diff
@@ -15,16 +16,30 @@ import { handleUserConfig, onConnectionLost, requestMiddleware } from "./middlew
...
+} else {
+  proxyApp = httpProxy.createProxyServer({
+    autoRewrite: true,
+    agent: new http.Agent({
+      keepAlive: true,
+      keepAliveMsecs: 5000,
+    }),
+  });
+  if (isHttpsUrl(SWA_CLI_API_URI())) {
+    logger.error(`Your connection is not secure. Please start the CLI using the flag --ssl. Exiting`, true);
```
manekinekko, line 41: We are introducing a breaking change. Please add this new behavior to the docs.

src/msha/server.ts
```diff
@@ -40,6 +55,19 @@ const httpsServerOptions: Pick<https.ServerOptions, "cert" | "key"> | null =
...
+  }
+
+  try {
+    const uri = new URL(url);
+    return uri.protocol.startsWith("https");
+  } catch {
+    return false;
+  }
+}
+
```
manekinekko, line 70: move this helper function under `core/...`

manekinekko (changes requested):

src/msha/server.ts
```diff
@@ -40,6 +55,19 @@ const httpsServerOptions: Pick<https.ServerOptions, "cert" | "key"> | null =
       }
     : null;
 
+function isHttpsUrl(url: string | undefined) {
```
sulabh-msft, line 58: Also can we add test for this method if it moving to core.

src/core/utils/net.spec.ts
```diff
@@ -310,4 +310,13 @@ describe("net utilities", () => {
...
     });
   });
+
+  describe("isHttpsUrl()", () => {
+    it("https url should be valid", () => {
+      expect(isHttpsUrl("https://foo.com")).toBeTrue();
+    });
+    it("http url should be invalid", () => {
+      expect(isHttpsUrl("http://foo.com")).toBeFalse();
+    });
```
manekinekko, line 320: please also add invalid values to make sure the code won't throw unexpected errors.

manekinekko (changes requested):

docs/www/docs/cli/swa-start.md
```diff
@@ -160,6 +160,10 @@ Connect both front-end and the API to running development server
 swa start http://localhost:3000 --api-devserver-url http://localhost:7071
 ```
 
+:::info Note
```
sulabh-msft, line 163: just for my understanding how will this section be rendered ?

src/core/utils/net.spec.ts
sulabh-msft: +1
Scenarios:
1. Null url
2. invalid url ( e..g random string  .. xysaa)
3. url with caps protocol - HTTPS4.

sulabh-msft (approved):

rupareddy5-21:
Fixes #610

manekinekko:
Great job @rupareddy5-21 🎉

PR #653, opened 2023-02-10
feat: telemetry collection basic changes

rupareddy5-21:
Used AppInisghts for javascript SDK and added a basic telemetry reporter class which is used to send event and error data to an AppInisghts instance. When this reporter is called it lazy loads the appinsights client and sends data as a batch to appinisghts instance. Until the client is loaded event and exception data are stored in queues. Collection of required telemetry data from CLI code will be taken care of in next PR.

### DIFF
--- a/src/core/telemetry/baseTelemetryReporter.ts
+++ b/src/core/telemetry/baseTelemetryReporter.ts
@@ -1,0 +1,71 @@
+import type { TelemetryEventMeasurements, RawTelemetryEventProperties, TelemetryEventProperties } from "./telemetryReporterTypes";
+import { BaseTelemetrySender } from "./baseTelemetrySender";
+
+export interface SenderData {
+  properties?: RawTelemetryEventProperties;
+  measurements?: TelemetryEventMeasurements;
+}
+
+export class BaseTelemetryReporter {
+  constructor(private telemetrySender: BaseTelemetrySender) {
+    this.telemetrySender.instantiateSender();
+  }
+  /**
+   * Internal function which logs telemetry events and takes extra options.
+   * @param eventName The name of the event
+   * @param properties The properties of the event
+   * @param measurements The measurements (numeric values) to send with the event
+   */
+  private internalSendTelemetryEvent(
+    eventName: string,
+    properties: TelemetryEventProperties | undefined,
+    measurements: TelemetryEventMeasurements | undefined
+  ): void {
+    this.telemetrySender.sendEventData(eventName, { properties, measurements });
+  }
+
+  /**
+   * Given an event name, some properties, and measurements sends a telemetry event.
+   * Properties are sanitized on best-effort basis to remove sensitive data prior to sending.
+   * @param eventName The name of the event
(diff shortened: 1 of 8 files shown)


src/core/telemetry/telemetryReporter.ts
```diff
@@ -0,0 +1,80 @@
...
+    },
+  };
+
+  return telemetryClient;
+};
+
+export default class TelemetryReporter extends BaseTelemetryReporter {
+  constructor(key: string | undefined) {
+    if (key) {
+      let clientFactory = (key: string) => appInsightsClientFactory(key);
```
sulabh-msft, line 75: `let` -> `const` ?

src/core/telemetry/telemetryReporter.ts
```diff
@@ -0,0 +1,80 @@
...
+        throw new Error("Failed to flush app insights!\n" + e.message);
+      }
+    },
+  };
+
+  return telemetryClient;
+};
+
+export default class TelemetryReporter extends BaseTelemetryReporter {
+  constructor(key: string | undefined) {
```
sulabh-msft, line 73: in case key is undefined.. what are we doing ?

src/core/telemetry/baseTelemetryReporter.ts
```diff
@@ -0,0 +1,71 @@
...
+  constructor(private telemetrySender: BaseTelemetrySender) {
+    this.telemetrySender.instantiateSender();
+  }
+  /**
+   * Internal function which logs telemetry events and takes extra options.
+   * @param eventName The name of the event
+   * @param properties The properties of the event
+   * @param measurements The measurements (numeric values) to send with the event
+   */
+  private internalSendTelemetryEvent(
```
sulabh-msft, line 19: shall we move the private method below the public method ?

src/core/telemetry/baseTelemetrySender.ts
```diff
@@ -0,0 +1,107 @@
...
+      await this._telemetryClient.flush();
+      this._telemetryClient = undefined;
+    }
+    return;
+  }
+
+  /**
+   * Flushes the queued events that existed before the client was instantiated
+   */
+  private _flushQueues(): void {
```
sulabh-msft, line 81: shall we move this method below the public methods?

src/core/telemetry/baseTelemetrySender.ts
```diff
@@ -0,0 +1,107 @@
...
+    this._telemetryClient.logException(exception, data);
+  }
+
+  /**
+   * Flushes the buffered telemetry data
+   */
+  async flush(): Promise<void> {
+    if (this._telemetryClient) {
+      await this._telemetryClient.flush();
+      this._telemetryClient = undefined;
```
sulabh-msft, line 73: are you sure that we need to set the telemetryclient to undefined after flush? can we call it only once?

sulabh-msft:
@rupareddy5-21  Can you please update the description of the PR ?

sulabh-msft (approved):

PR #656, opened 2023-02-17
feat(telemetry): added new flags and env variable for opting out

rupareddy5-21:
Added commands to disable/enable telemetry capturing. Command would look like `swa telemetry --disable/enable`.
Also added an environment variable `SWA_DISABLE_TELEMETRY` where users can set the value to true/false.
Users can check the status of telemetry capturing using the command `swa telemetry --status`.
Only one of these flags can be used at a time.

### DIFF
--- a/src/cli/commands/telemetry/register.ts
+++ b/src/cli/commands/telemetry/register.ts
@@ -6,7 +6,7 @@ export default function registerCommand(program: Command) {
   program
     .command("telemetry")
     .usage("[options]")
-    .description("telemetry")
+    .description("enable/disable SWA CLI telemetry capturing")
     .option("-dt, --disable", "opt out of telemetry", undefined)
     .option("-et, --enable", "opt in telemetry", undefined)
     .option("-st, --status", "status of telemetry", undefined)
--- a/src/cli/commands/telemetry/register.ts
+++ b/src/cli/commands/telemetry/register.ts
@@ -1,0 +1,17 @@
+import { Command } from "commander";
+import { configureOptions } from "../../../core";
+import { telemetry } from "./telemetry";
+
+export default function registerCommand(program: Command) {
+  program
+    .command("telemetry")
+    .usage("[options]")
+    .description("telemetry")
+    .option("-dt, --disable", "opt out of telemetry", undefined)
+    .option("-et, --enable", "opt in telemetry", undefined)
+    .option("-st, --status", "status of telemetry", undefined)
+    .action(async (_options: SWACLIConfig, command: Command) => {
+      const options = await configureOptions(undefined, command.optsWithGlobals(), command, "telemetry");
+      await telemetry(options);
+    });
+}
--- a/src/cli/commands/telemetry/register.ts
+++ b/src/cli/commands/telemetry/register.ts
@@ -6,10 +6,10 @@ export default function registerCommand(program: Command) {
   program
     .command("telemetry")
     .usage("[options]")
-    .description("enable/disable SWA CLI telemetry capturing")
-    .option("-dt, --disable", "opt out of telemetry", undefined)
-    .option("-et, --enable", "opt in telemetry", undefined)
-    .option("-st, --status", "status of telemetry", undefined)
+    .description("manage SWA CLI telemetry capturing")
+    .option("-dt, --disable", "opt out of telemetry capturing", undefined)
+    .option("-et, --enable", "opt in telemetry capturing", undefined)
+    .option("-st, --status", "status of telemetry capturing", undefined)
     .action(async (_options: SWACLIConfig, command: Command) => {
       const options = await configureOptions(undefined, command.optsWithGlobals(), command, "telemetry");
       await telemetry(options);
--- a/src/cli/commands/telemetry/register.ts
+++ b/src/cli/commands/telemetry/register.ts
@@ -13,5 +13,20 @@ export default function registerCommand(program: Command) {
     .action(async (_options: SWACLIConfig, command: Command) => {
       const options = await configureOptions(undefined, command.optsWithGlobals(), command, "telemetry");
       await telemetry(options);
-    });
+    })
+    .addHelpText(
+      "after",
+      `
+Examples:
+
+opt out of telemetry capturing
+swa telemetry --disable
+
+opt in telemetry capturing
+swa telemetry --enable
+
+check the status of telemetry capturing
+saw telemetry --status
+      `
+    );
 }
--- a/src/cli/commands/telemetry/telemetry.spec.ts
+++ b/src/cli/commands/telemetry/telemetry.spec.ts
@@ -1,0 +1,94 @@
+import fs from "fs";
+import mockFs from "mock-fs";
+import os from "os";
+import path from "path";
+import { logger } from "../../../core";
+import { ENV_FILENAME } from "../../../core/constants";
+import { telemetry } from "./telemetry";
+
+jest.mock("../../../core/utils/logger", () => {
+  return {
+    logger: {
+      error: jest.fn(),
+      log: jest.fn(),
+      warn: jest.fn(),
+      silly: jest.fn(),
+    },
+    logGiHubIssueMessageAndExit: jest.fn(),
+  };
+});
+
+describe("swa telemetry", () => {
+  afterEach(() => {
+    mockFs.restore();
+  });
+
+  it("Telemetry should be enabled by default", async () => {
+    mockFs();
+    await telemetry({ status: true });
+    expect(logger.log).toHaveBeenNthCalledWith(1, "Telemetry capturing is enabled.");
+  });
+
+  it("should store disable telemetry setting in .env file", async () => {
+    await telemetry({ disable: true });
+    const envFile = path.join(os.homedir(), ".swa", ENV_FILENAME);
+    const envFileContent = fs.readFileSync(envFile, "utf-8");
+
+    expect(envFileContent).toMatchSnapshot(`
+        "SWA_DISABLE_TELEMETRY=true"
+        `);
+  });
+
+  it("should log the telemetry setting as disabled", async () => {
+    await telemetry({ status: true });
+    expect(logger.log).toHaveBeenNthCalledWith(3, "Telemetry capturing is disabled.");
+  });
+
+  it("should store enable telemetry setting in .env file", async () => {
+    await telemetry({ enable: true });
+    const envFile = path.join(os.homedir(), ".swa", ENV_FILENAME);
+    const envFileContent = fs.readFileSync(envFile, "utf-8");
+
+    expect(envFileContent).toMatchSnapshot(`
+        "SWA_DISABLE_TELEMETRY=false"
+        `);
+  });
+
+  it("should log the telemetry setting as enabled", async () => {
+    await telemetry({ status: true });
+    expect(logger.log).toHaveBeenNthCalledWith(5, "Telemetry capturing is enabled.");
+  });
+
+  it("should disable telemetry when both --enable and --disable flags are used", async () => {
+    await telemetry({ disable: true, enable: true });
+    const envFile = path.join(os.homedir(), ".swa", ENV_FILENAME);
+    const envFileContent = fs.readFileSync(envFile, "utf-8");
+
+    expect(envFileContent).toMatchSnapshot(`
+        "SWA_DISABLE_TELEMETRY=true"
+        `);
+  });
+
+  it("should log the telemetry setting as disabled", async () => {
+    await telemetry({ status: true });
+    expect(logger.log).toHaveBeenNthCalledWith(7, "Telemetry capturing is disabled.");
+  });
+
+  it("should warn when both --enable and --status flags are used", async () => {
+    mockFs();
+    await telemetry({ enable: true, status: true });
+    expect(logger.warn).toHaveBeenNthCalledWith(1, "The flags --disable and --enable can't be used alongside the flag --status!");
+  });
+
+  it("should warn when both --disable and --status flags are used", async () => {
+    mockFs();
+    await telemetry({ disable: true, status: true });
+    expect(logger.warn).toHaveBeenNthCalledWith(1, "The flags --disable and --enable can't be used alongside the flag --status!");
+  });
+
+  it("should warn when all 3 --disable, --enable and --status flags are used", async () => {
+    mockFs();
+    await telemetry({ enable: true, disable: true, status: true });
+    expect(logger.warn).toHaveBeenNthCalledWith(1, "The flags --disable and --enable can't be used alongside the flag --status!");
+  });
+});
--- a/src/cli/commands/telemetry/telemetry.spec.ts
+++ b/src/cli/commands/telemetry/telemetry.spec.ts
@@ -35,7 +35,7 @@ describe("swa telemetry", () => {
     const envFileContent = fs.readFileSync(envFile, "utf-8");
 
     expect(envFileContent).toMatchSnapshot(`
-        "SWA_DISABLE_TELEMETRY=true"
+        "SWA_CLI_CAPTURE_TELEMETRY=false"
         `);
   });
 
@@ -50,7 +50,7 @@ describe("swa telemetry", () => {
     const envFileContent = fs.readFileSync(envFile, "utf-8");
 
     expect(envFileContent).toMatchSnapshot(`
-        "SWA_DISABLE_TELEMETRY=false"
+        "SWA_CLI_CAPTURE_TELEMETRY=true"
         `);
   });
 
@@ -59,36 +59,23 @@ describe("swa telemetry", () => {
     expect(logger.log).toHaveBeenNthCalledWith(5, "Telemetry capturing is enabled.");
   });
 
-  it("should disable telemetry when both --enable and --disable flags are used", async () => {
+  it("should warn when both --enable and --disable flags are used", async () => {
     await telemetry({ disable: true, enable: true });
-    const envFile = path.join(os.homedir(), ".swa", ENV_FILENAME);
-    const envFileContent = fs.readFileSync(envFile, "utf-8");
-
-    expect(envFileContent).toMatchSnapshot(`
-        "SWA_DISABLE_TELEMETRY=true"
-        `);
-  });
-
-  it("should log the telemetry setting as disabled", async () => {
-    await telemetry({ status: true });
-    expect(logger.log).toHaveBeenNthCalledWith(7, "Telemetry capturing is disabled.");
+    expect(logger.warn).toHaveBeenNthCalledWith(1, "Using the flags --disable and --enable together is not supported!");
   });
 
   it("should warn when both --enable and --status flags are used", async () => {
-    mockFs();
     await telemetry({ enable: true, status: true });
-    expect(logger.warn).toHaveBeenNthCalledWith(1, "The flag --status can't be used alongside the flags --disable and --enable!");
+    expect(logger.warn).toHaveBeenNthCalledWith(2, "The flag --status can't be used alongside the flags --disable and --enable!");
   });
 
   it("should warn when both --disable and --status flags are used", async () => {
-    mockFs();
     await telemetry({ disable: true, status: true });
-    expect(logger.warn).toHaveBeenNthCalledWith(1, "The flag --status can't be used alongside the flags --disable and --enable!");
+    expect(logger.warn).toHaveBeenNthCalledWith(3, "The flag --status can't be used alongside the flags --disable and --enable!");
   });
 
   it("should warn when all 3 --disable, --enable and --status flags are used", async () => {
-    mockFs();
     await telemetry({ enable: true, disable: true, status: true });
-    expect(logger.warn).toHaveBeenNthCalledWith(1, "The flag --status can't be used alongside the flags --disable and --enable!");
+    expect(logger.warn).toHaveBeenNthCalledWith(4, "The flag --status can't be used alongside the flags --disable and --enable!");
   });
 });
--- a/src/cli/commands/telemetry/telemetry.spec.ts
+++ b/src/cli/commands/telemetry/telemetry.spec.ts
@@ -77,18 +77,18 @@ describe("swa telemetry", () => {
   it("should warn when both --enable and --status flags are used", async () => {
     mockFs();
     await telemetry({ enable: true, status: true });
-    expect(logger.warn).toHaveBeenNthCalledWith(1, "The flags --disable and --enable can't be used alongside the flag --status!");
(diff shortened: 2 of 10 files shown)


sgollapudi77:
On a second thought, if we're just gonna have two flags, one to `--enable/disable` and one for `--status` why do we need to have a separate sub command? Can't we just have it as two flags under the main 'swa' command like '--telemetry<bool>' (takes bool as input for enabling/disabling the telemetry. Ex: `swa --telemetry=false`) and another flag like `swa --telemetry-status` to show the status.
What do you think about this @sulabh-msft ??

Reshmi-Sriram:
> On a second thought, if we're just gonna have two flags, one to `--enable/disable` and one for `--status` why do we need to have a separate sub command? Can't we just have it as two flags under the main 'swa' command like '--telemetry' (takes bool as input for enabling/disabling the telemetry. Ex: `swa --telemetry=false`) and another flag like `swa --telemetry-status` to show the status. What do you think about this @sulabh-msft and @manekinekko ?

@rupareddy5-21 and I discussed about this. While it is an easier implementation, it has a couple of catches. 

- if it is limited to just a global flag usage, i.e. `swa --telemetry=false`, we already have `swa` itself as an executable command and we'll have to make code changes to ensure if `--telemetry` is used then we shouldn't trigger `swa`.
- If it is a part of a string of other flags, recognizing it as a separate globally actionable flag will be tricky
- Also, since this is regarding telemetry capturing, this flag will have to take priority over other flags and thus if it is present, we need to add rules to not capture telemetry even for the command when it is called out (e.g. `swa start ./dist --api-location=api --telemetry=false` needs us to not capture the `swa start` in the telemetry as well

Due to all this, we thought having a separate command group will be the cleanest and most user-friendly way of doing this.

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,57 @@
+import { logger } from "../../../core";
+import path from "path";
+import dotenv from "dotenv";
+import { ENV_FILENAME } from "../../../core/constants";
+import { updateGitIgnore } from "../../../core/git";
+import { existsSync, promises as fsPromises } from "fs";
+const { readFile, writeFile } = fsPromises;
+
+export async function telemetry(options: SWACLIConfig) {
+  let { disable, enable, status } = options;
```
sulabh-msft, line 10: `let` -> `const`

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,57 @@
+import { logger } from "../../../core";
```
sulabh-msft, line 1: shall we move the logger import after `import dotenv from "dotenv";`.. Idea is external ones we keep them together.. while internal relative references we keep together

src/cli/commands/telemetry/register.ts
```diff
@@ -0,0 +1,17 @@
+import { Command } from "commander";
+import { configureOptions } from "../../../core";
+import { telemetry } from "./telemetry";
+
+export default function registerCommand(program: Command) {
+  program
+    .command("telemetry")
+    .usage("[options]")
+    .description("enable/disable SWA CLI telemetry capturing")
+    .option("-dt, --disable", "opt out of telemetry", undefined)
```
sulabh-msft, line 10: `manage SWA CLI telemetry capturing` ?

sulabh-msft, line 10: opt out of telemetry capturing

sulabh-msft, line 10: any reason for using `-dt` two characters in the option? why not we are using `-d`, `-e` etc here?

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+import dotenv from "dotenv";
+import { ENV_FILENAME } from "../../../core/constants";
+import { updateGitIgnore } from "../../../core/git";
+import { existsSync, promises as fsPromises } from "fs";
+const { readFile, writeFile } = fsPromises;
+
+export async function telemetry(options: SWACLIConfig) {
+  let { disable, enable, status } = options;
+  const envFile = path.join(process.cwd(), ENV_FILENAME);
+  const envFileExists = existsSync(envFile);
```
sulabh-msft, line 12: needed to create a variable we are just using it in the next line.. shall we inline it ?

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+import path from "path";
+import dotenv from "dotenv";
+import { ENV_FILENAME } from "../../../core/constants";
+import { updateGitIgnore } from "../../../core/git";
+import { existsSync, promises as fsPromises } from "fs";
+const { readFile, writeFile } = fsPromises;
+
+export async function telemetry(options: SWACLIConfig) {
+  let { disable, enable, status } = options;
+  const envFile = path.join(process.cwd(), ENV_FILENAME);
```
sulabh-msft, line 11: Check if we can refactor and put the ENV_FILENAME reading and writing to a separate helper file and use it in login/telemetry flow

src/cli/commands/telemetry/register.ts
rupareddy5-21: -d, -e and -s are already in use for other swa commands

src/swa.d.ts
```diff
@@ -183,6 +186,11 @@ declare type SWACLISharedLoginOptions = LoginDetails & {
 
 declare type SWACLILoginOptions = SWACLISharedLoginOptions;
 
+declare type SWACLITelemetryOptions = {
+  disable?: boolean;
+  enable?: boolean;
+  status?: boolean;
```
sulabh-msft, line 192: Q: what is the experience in case we don't specify all these options will we get the capture telemetry status ?

src/swa.d.ts
```diff
@@ -71,6 +71,9 @@ declare interface SWACLIEnv extends StaticSiteClientEnv {
   AZURE_TENANT_ID?: string;
   AZURE_CLIENT_ID?: string;
   AZURE_CLIENT_SECRET?: string;
+
+  // swa telemetry
+  SWA_DISABLE_TELEMETRY?: string;
```
sulabh-msft, line 76: SWA_DISABLE_TELEMETRY -> `SWA_CAPTURE_TELEMETRY` ?

src/core/telemetry/utils.ts
```diff
@@ -5,11 +5,15 @@ import TelemetryReporter from "./telemetryReporter";
...
+  if (process.env.SWA_DISABLE_TELEMETRY == "false") {
+    const reporter: TelemetryReporter = new TelemetryReporter(aiKey);
+    reporter.sendTelemetryEvent(event, properties, measurements);
+  }
 }
 
 export function collectTelemetryException(exception: Error, properties?: TelemetryEventProperties, measurements?: TelemetryEventMeasurements) {
-  const reporter: TelemetryReporter = new TelemetryReporter(process.env.APPLICATIONINSIGHTS_CONNECTION_STRING);
-  reporter.sendTelemetryException(exception, properties, measurements);
+  if (process.env.SWA_DISABLE_TELEMETRY == "false") {
```
sulabh-msft, line 15: Q: will this environment variable be set/unset if we are setting the telemetry flag through the swa command.

sgollapudi77:
> 
> @rupareddy5-21 and I discussed about this. While it is an easier implementation, it has a couple of catches.
> 
> * if it is limited to just a global flag usage, i.e. `swa --telemetry=false`, we already have `swa` itself as an executable command and we'll have to make code changes to ensure if `--telemetry` is used then we shouldn't trigger `swa`.
> * If it is a part of a string of other flags, recognizing it as a separate globally actionable flag will be tricky
> * Also, since this is regarding telemetry capturing, this flag will have to take priority over other flags and thus if it is present, we need to add rules to not capture telemetry even for the command when it is called out (e.g. `swa start ./dist --api-location=api --telemetry=false` needs us to not capture the `swa start` in the telemetry as well
> 
> Due to all this, we thought having a separate command group will be the cleanest and most user-friendly way of doing this.

Replying to your points
1. 'swa' itself is an executable but invoking `swa --telemetry` will be nothing different than invoking `swa --version` which is a global flag. 
2. This can be avoided by checking the other existing flags and renaming appropriately.
3. We can prevent this from happening when parsing the options for subcommand. Basically, we can prevent subcommand from invoking this.

Upon saying this, I think since this is a big change it'd be better to have a discussion before merging this.

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+  const envFileExists = existsSync(envFile);
+  const envFileContent = envFileExists ? await readFile(envFile, "utf8") : "";
+  const buf = Buffer.from(envFileContent);
+
+  // in case the .env file format changes in the future, we can use the following to parse the file
+  const config = dotenv.parse(buf);
+  const oldEnvFileLines = Object.keys(config).map((key) => `${key}=${config[key]}`);
+  const newEnvFileLines = [];
+
+  if ((disable || enable) && !status) {
```
sulabh-msft, line 21: so much of check.. for enable/disable.. shall we think of using `-capture=true/false` and -status  as Vamsi is suggesting

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+    let disableTelemetryStatus = undefined;
+    if (disable && enable == undefined) {
+      disableTelemetryStatus = "true";
```
sulabh-msft, line 24: Default value of telemetry capture is `true` right.. so if customer is not opting out .. telemetry capturing will be enabled by defauly.

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+    } else if (disable == undefined && enable) {
+      disableTelemetryStatus = "false";
+    } else {
+      logger.warn("The flags --disable and --enable can't be used at the same time!");
+    }
+
+    let entry = `SWA_DISABLE_TELEMETRY=${disableTelemetryStatus}`;
```
sulabh-msft, line 31: let -> `const`

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    if (disableTelemetryStatus) {
+      if (!envFileContent.includes("SWA_DISABLE_TELEMETRY")) {
+        newEnvFileLines.push(entry);
+      } else {
+        const index = oldEnvFileLines.indexOf(`SWA_DISABLE_TELEMETRY=${process.env.SWA_DISABLE_TELEMETRY}`);
```
sulabh-msft, line 37: so previous state can be true or false.. how are we handling it here?

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+      }
+    }
+
+    // write file if we have at least one new env line
+    if (newEnvFileLines.length > 0) {
+      const envFileContentWithProjectDetails = [...oldEnvFileLines, ...newEnvFileLines].join("\n");
+      await writeFile(envFile, envFileContentWithProjectDetails);
+
+      await updateGitIgnore(ENV_FILENAME);
+    }
```
sulabh-msft, line 49: move to a helper method and  use it in both places.

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    if (disableTelemetryStatus) {
+      if (!envFileContent.includes("SWA_DISABLE_TELEMETRY")) {
+        newEnvFileLines.push(entry);
+      } else {
+        const index = oldEnvFileLines.indexOf(`SWA_DISABLE_TELEMETRY=${process.env.SWA_DISABLE_TELEMETRY}`);
+        oldEnvFileLines.splice(index, 1);
+        newEnvFileLines.push(entry);
+      }
+    }
```
sulabh-msft, line 41: should we also inform the users that we are saving the settings in the ENV_FILENAME

src/cli/commands/telemetry/register.ts
sulabh-msft: I see so these short forms are not within a command group its global

src/swa.d.ts
rupareddy5-21: if we dont specify one of the flags, just `swa telemetry`, then nothing happens, wont show the status.

src/core/telemetry/utils.ts
rupareddy5-21: yes

src/cli/commands/telemetry/telemetry.ts
rupareddy5-21: whatever the previous state might be, it is stored in process.env.SWA_DISABLE_TELEMETRY

src/cli/commands/telemetry/telemetry.ts
rupareddy5-21: please check latest changes, updating the .env file is different in both the cases and we are writing to different .env files.

src/swa.d.ts
rupareddy5-21: As this is something users can set/reset to disable the telemetry capturing, i think SWA_DISABLE_TELEMETRY makes more sense no?

src/core/telemetry/utils.ts
```diff
@@ -5,11 +5,15 @@ import TelemetryReporter from "./telemetryReporter";
 import type { TelemetryEventMeasurements, TelemetryEventProperties } from "./telemetryReporterTypes";
 
 export function collectTelemetryEvent(event: string, properties?: TelemetryEventProperties, measurements?: TelemetryEventMeasurements) {
-  const reporter: TelemetryReporter = new TelemetryReporter(process.env.APPLICATIONINSIGHTS_CONNECTION_STRING);
-  reporter.sendTelemetryEvent(event, properties, measurements);
+  if (process.env.SWA_DISABLE_TELEMETRY == "false") {
+    const reporter: TelemetryReporter = new TelemetryReporter(aiKey);
```
sulabh-msft, line 9: where are we setting `aiKey` ?

src/swa.d.ts
sulabh-msft: ok.. then lets scope the string value to two values then ? ( true/false )

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    // write file if we have at least one new env line
+    if (newEnvFileLines.length > 0) {
+      const envFileContentWithProjectDetails = [...oldEnvFileLines, ...newEnvFileLines].join("\n");
+      await writeFile(envFile, envFileContentWithProjectDetails);
+      logger.log(chalk.green(`✔ Saved Telemetry setting in ${ENV_FILENAME} file at ${path.join(os.homedir(), ".swa")}`));
+    }
+  } else if (!disable && !enable && status) {
+    if (config["SWA_DISABLE_TELEMETRY"] == "true") {
```
sulabh-msft, line 50: does it do ignore case comparison ? `True` ?

sulabh-msft (approved):
Please get the PR reviewed by Vamsi as well.

sgollapudi77:
> Please get the PR reviewed by Vamsi as well.

Rupa, can you resolve the previous comments?

sgollapudi77 (changes requested):

src/core/telemetry/utils.ts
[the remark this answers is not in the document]
rupareddy5-21: we now get the env variable flag from .env file in .swa folder and it's a string, not a boolean value

src/cli/commands/telemetry/telemetry.ts
rupareddy5-21: added a helper method to read the env file.

src/swa.d.ts
rupareddy5-21: users can change the env variable value right, how can we control that? added case insensitive comparision in the code.

src/cli/commands/telemetry/telemetry.ts
rupareddy5-21: added a method to read file

src/cli/commands/telemetry/telemetry.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+        `);
+  });
+
+  it("should log the telemetry setting as enabled", async () => {
+    await telemetry({ status: true });
+    expect(logger.log).toHaveBeenNthCalledWith(5, "Telemetry capturing is enabled.");
+  });
+
+  it("should disable telemetry when both --enable and --disable flags are used", async () => {
+    await telemetry({ disable: true, enable: true });
```
sulabh-msft, line 63: we should call it as unsupported as we shouldn't allow passing those flags together.

src/swa.d.ts
sgollapudi77: > SWA_DISABLE_TELEMETRY -> `SWA_CAPTURE_TELEMETRY` ? in this case.. they will set it to false.. default value of it will be true

CAPTURE_TELEMETRY makes sense. I agree with Sulabh.

rupareddy5-21: bu all the other CLIs also have the env variable name as DISABLE_TELEMETRY

src/cli/commands/telemetry/telemetry.spec.ts
rupareddy5-21: @Reshmi-Sriram thoughts on using --enable and --disable together?

sulabh-msft: We should keep it simple.
just support passing only one option here:

--enable
--disable
--status

any combination of these will not make sense and in my opinion should not be allowed and appropriate message can be given to the customer.

src/swa.d.ts
sulabh-msft: The reason why I gave the suggestion of using SWA_CAPTURE_TELEMETRY was every where in the docs/command help text we are talking about... Capture telemetry ..  it was more natural to have command in that format.

https://learn.microsoft.com/en-us/cli/azure/azure-cli-configuration

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    // write file if we have at least one new env line
+    if (newEnvFileLines.length > 0) {
+      const envFileContentWithProjectDetails = [...oldEnvFileLines, ...newEnvFileLines].join("\n");
+      const envFile = path.join(os.homedir(), ".swa", ENV_FILENAME);
+      await writeFile(envFile, envFileContentWithProjectDetails);
+      logger.log(chalk.green(`✔ Saved Telemetry setting in ${ENV_FILENAME} file at ${path.join(os.homedir(), ".swa")}`));
+    }
+  } else if (!disable && !enable && status) {
+    if (config["SWA_DISABLE_TELEMETRY"] && config["SWA_DISABLE_TELEMETRY"].toLowerCase() === "true") {
```
sulabh-msft, line 46: is it possible to read the `config["SWA_DISABLE_TELEMETRY"]` in a variable and use in if/else statements.

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,54 @@
+import chalk from "chalk";
+import os from "os";
+import path from "path";
+import { logger } from "../../../core";
+import { readCLIEnvFile } from "../../../core";
+import { ENV_FILENAME } from "../../../core/constants";
+import { promises as fsPromises } from "fs";
```
sulabh-msft, line 7: nit: move promises import to `path import`

src/core/utils/file.ts
```diff
@@ -1,7 +1,11 @@
-import { promises as fs } from "fs";
+import dotenv from "dotenv";
+import { existsSync, promises as fs } from "fs";
+import os from "os";
 import path from "path";
 import { logger } from "./logger";
+import { ENV_FILENAME } from "./../constants";
 import { stripJsonComments } from "./strings";
+const { readFile } = fs;
```
sulabh-msft, line 9: nit: move readFile to the following import

``import { existsSync, promises as fs } from "fs";``


``import { existsSync, readFile, promises as fs } from "fs";``

src/swa.d.ts
sulabh-msft: Also noticed  we are using .. `SWA_CLI..` as the prefix for the env variables. if we are deciding on any name lets include `SWA_CLI` as the prefix for sure

src/core/telemetry/utils.ts
```diff
@@ -3,13 +3,21 @@ dotenv.config();
...
+    reporter.sendTelemetryEvent(event, properties, measurements);
+  }
 }
 
-export function collectTelemetryException(exception: Error, properties?: TelemetryEventProperties, measurements?: TelemetryEventMeasurements) {
-  const reporter: TelemetryReporter = new TelemetryReporter(process.env.APPLICATIONINSIGHTS_CONNECTION_STRING);
-  reporter.sendTelemetryException(exception, properties, measurements);
+export async function collectTelemetryException(exception: Error, properties?: TelemetryEventProperties, measurements?: TelemetryEventMeasurements) {
+  const config = await readCLIEnvFile();
+  if (config["SWA_DISABLE_TELEMETRY"].toLowerCase() === "false") {
```
sulabh-msft, line 19: move to a helper method?.. sort of repeated in each of these methods.

src/cli/commands/telemetry/telemetry.ts
```diff
@@ -0,0 +1,54 @@
...
+        const index = oldEnvFileLines.indexOf(`SWA_DISABLE_TELEMETRY=${config["SWA_DISABLE_TELEMETRY"]}`);
+        oldEnvFileLines.splice(index, 1);
+        newEnvFileLines.push(entry);
+      }
+    }
+
+    // write file if we have at least one new env line
+    if (newEnvFileLines.length > 0) {
+      const envFileContentWithProjectDetails = [...oldEnvFileLines, ...newEnvFileLines].join("\n");
+      const envFile = path.join(os.homedir(), ".swa", ENV_FILENAME);
```
sulabh-msft, line 41: for reducing the duplication, we can save `path.join(os.homedir(), ".swa")  in a varialble and use it here as well as while logging below

src/cli/commands/telemetry/telemetry.spec.ts
```diff
@@ -0,0 +1,94 @@
+import fs from "fs";
```
sulabh-msft, line 1: Thanks! for adding the tests. Appreciate it.

src/core/utils/file.ts
rupareddy5-21: we cant do this as we get readfile from {promises} not from "fs";

sgollapudi77 (approved):

PR #662, opened 2023-03-01
feat: dab integration

vivekjilla:


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@azure/static-web-apps-cli",
-  "version": "1.0.6",
```
sulabh-msft, line 3: Lets validate it once.. should it be alpa?

sgollapudi77: yes, modified it now.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@azure/static-web-apps-cli",
-  "version": "1.0.7-alpha",
```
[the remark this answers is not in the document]
sgollapudi77: done

src/core/dataApiBuilder/dab.ts
```diff
@@ -101,7 +100,8 @@ async function getReleaseDataApiBuilderMetadata(): Promise<{ releaseMetadata: Da
   const responseMetadata = (await response.json()) as DataApiBuilderReleaseMetadata;
 
   if (Array.isArray(responseMetadata)) {
-    const releaseMetadata = responseMetadata.find((c) => c.releaseType === DATA_API_BUILDER_RELEASE_TAG);
+    // const releaseMetadata = responseMetadata.find((c) => c.version === DATA_API_BUILDER_LATEST_TAG); // If we want to proceed with downloading the latest tag
```
vivekjilla, line 103: should we also continue to check for releaseType as well, no?

sgollapudi77: releaseType is released for all the released versions, so it can be added as an additional check but not useful imo.

aaronpowell:
Can we make sure that https://github.com/Azure/static-web-apps-cli/blob/main/schema/swa-cli.config.schema.json also gets updated with the new config property? It looks like the config supports it but it's not in the schema which means it doesn't show up properly when authoring the config.

thomasgauvin:
> Can we make sure that https://github.com/Azure/static-web-apps-cli/blob/main/schema/swa-cli.config.schema.json also gets updated with the new config property? It looks like the config supports it but it's not in the schema which means it doesn't show up properly when authoring the config.

These schema changes are available in this PR: https://github.com/Azure/static-web-apps-cli/pull/664

aaronpowell:
> > Can we make sure that https://github.com/Azure/static-web-apps-cli/blob/main/schema/swa-cli.config.schema.json also gets updated with the new config property? It looks like the config supports it but it's not in the schema which means it doesn't show up properly when authoring the config.
> 
> These schema changes are available in this PR: #664.

My bad - I just assumed everything was in this PR

sgollapudi77:
@all-contributors add @vivekjilla code

sgollapudi77:
@all-contributors add @vivekjilla code

PR #664, opened 2023-03-06
Add support for custom data api location for the deploy command

thomasgauvin:
Data API Location can be specified via cli arg, config file parameter, workflow file parameter

### DIFF
--- a/src/cli/commands/deploy/deploy.ts
+++ b/src/cli/commands/deploy/deploy.ts
@@ -38,7 +38,7 @@ export async function deploy(options: SWACLIConfig) {
   if (dataApiLocation) {
     dataApiLocation = path.resolve(dataApiLocation);
     if (!fs.existsSync(dataApiLocation)) {
-      logger.error(`The provided API folder ${dataApiLocation} does not exist. Abort.`, true);
+      logger.error(`The provided Data API folder ${dataApiLocation} does not exist. Abort.`, true);
       return;
     } else {
       logger.log(`Deploying Data API from folder:`);
--- a/src/cli/commands/deploy/deploy.ts
+++ b/src/cli/commands/deploy/deploy.ts
@@ -37,6 +37,14 @@ export async function deploy(options: SWACLIConfig) {
   // make sure dataApiLocation is set
   if(dataApiLocation) {
     dataApiLocation = path.resolve(dataApiLocation);
+    if (!fs.existsSync(dataApiLocation)) {
+      logger.error(`The provided API folder ${dataApiLocation} does not exist. Abort.`, true);
+      return;
+    } else {
+      logger.log(`Deploying Data API from folder:`);
+      logger.log(`  ${chalk.green(dataApiLocation)}`);
+      logger.log(``);
+    }
   }
 
   // make sure outputLocation is set
--- a/src/cli/commands/deploy/deploy.ts
+++ b/src/cli/commands/deploy/deploy.ts
@@ -22,7 +22,7 @@ export async function deploy(options: SWACLIConfig) {
   const { SWA_CLI_DEPLOYMENT_TOKEN, SWA_CLI_DEBUG } = swaCLIEnv();
   const isVerboseEnabled = SWA_CLI_DEBUG === "silly";
 
-  let { appLocation, apiLocation, outputLocation, dryRun, deploymentToken, printToken, appName, swaConfigLocation, verbose } = options;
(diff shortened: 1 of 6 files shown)


src/cli/commands/deploy/deploy.ts
```diff
@@ -34,6 +34,11 @@ export async function deploy(options: SWACLIConfig) {
   // make sure appLocation is set
   appLocation = path.resolve(appLocation || process.cwd());
 
+  // make sure dataApiLocation is set
```
vivekjilla, line 37: I think we also use some default folder name for data-api-location. Do we also need to pass that folder name in default scenarios.

src/cli/commands/deploy/deploy.ts
```diff
@@ -34,6 +34,11 @@ export async function deploy(options: SWACLIConfig) {
...
+  if(dataApiLocation) {
```
sulabh-msft, line 38: nit: `if(dataApiLocation) {` -> `if (dataApiLocation) {`

src/cli/commands/deploy/deploy.ts
```diff
@@ -34,6 +34,11 @@ export async function deploy(options: SWACLIConfig) {
...
+    dataApiLocation = path.resolve(dataApiLocation);
```
sulabh-msft, line 39: should we be logging in the console about the dataApiLocation

src/cli/commands/deploy/deploy.ts
```diff
@@ -161,6 +166,7 @@ export async function deploy(options: SWACLIConfig) {
     appLocation,
     outputLocation: resolvedOutputLocation,
     apiLocation: resolvedApiLocation,
+    dataApiLocation
```
sulabh-msft, line 169: Any change do we need to do in the readWorkflowFile for enabling it to handle dataApiLocation ?

src/cli/commands/deploy/deploy.ts
thomasgauvin: By convention, if the folder `swa-db-connections` is in the root directory, it gets deployed (no need to pass default args to the StaticSitesClient)

src/cli/commands/deploy/deploy.ts
thomasgauvin, line 38:
```suggestion
  if (dataApiLocation) {
```

thomasgauvin: done! thanks

src/cli/commands/deploy/deploy.ts
thomasgauvin: Good catch thanks Sulabh, this & error logs/validation of path is missing, adding it

src/cli/commands/deploy/deploy.ts
thomasgauvin: Thanks Sulabh, I've adjusted this as well. I noticed that this feature currently broken while dev'ing and testing. I will let the bugfix take place in another PR.

thomasgauvin:
From my testing, SWA CLI currently doesn't work well with extracting workflow config from gh action workflow files, needs some bug fixing there since the values are extracted but not properly used. That is to be addressed in a separate PR

src/cli/commands/deploy/deploy.ts
sulabh-msft: yup make sense. @sgollapudi77 please create a tracking task if we don't have it already.

src/core/utils/workflow-config.ts
```diff
@@ -118,19 +119,24 @@ export function readWorkflowFile({ userWorkflowConfig }: { userWorkflowConfig?:
...
   // - output_location
+  // - data_api_location
 
   app_location = path.normalize(path.join(process.cwd(), app_location));
   if (typeof api_location !== "undefined") {
     api_location = path.normalize(path.join(process.cwd(), api_location || path.sep));
   }
   output_location = path.normalize(output_location);
   output_location = path.join(app_location, output_location);
+  if(typeof data_api_location !== "undefined") {
```
sulabh-msft, line 137:
```suggestion
  if (typeof data_api_location !== "undefined") {
```

sulabh-msft (approved):
Looks good. Left one nit comment.

src/cli/commands/deploy/deploy.ts
sgollapudi77: Added a task now Sulabh.

PR #675, opened 2023-03-16
fix: add tags for Staticsiteclient latest and dab

sgollapudi77:


sgollapudi77:
Not going to merge it in this iteration.

PR #678, opened 2023-03-20
feat(telemetry): changes to start collecting basic telemetry fields

rupareddy5-21:
OperationName field is set to the Name of the event in trackEvent method.
Pending fields - flagsUsed, InstallationId, errorMessage, response and error types.

### DIFF
--- a/src/cli/commands/build/build.ts
+++ b/src/cli/commands/build/build.ts
@@ -14,7 +14,7 @@ import { collectTelemetryEvent } from "../../../core/telemetry/utils";
 import { TELEMETRY_EVENTS } from "../../../core/constants";
 
 export async function build(options: SWACLIConfig) {
-  const start = new Date().getTime();
+  const cmdStartTime = new Date().getTime();
   const workflowConfig = readWorkflowFile();
 
   logger.silly({
@@ -107,10 +107,10 @@ export async function build(options: SWACLIConfig) {
     logger.log(`Building api with ${chalk.green(apiBuildCommand)} in ${chalk.dim(apiLocation)} ...`);
     runCommand(apiBuildCommand, apiLocation!);
   }
-  const end = new Date().getTime();
+  const cmdEndTime = new Date().getTime();
 
   collectTelemetryEvent(TELEMETRY_EVENTS.Build, {
-    duration: (end - start).toString(),
+    duration: (cmdEndTime - cmdStartTime).toString(),
   });
 }
 
--- a/src/cli/commands/build/build.ts
+++ b/src/cli/commands/build/build.ts
@@ -41,10 +41,9 @@ export async function build(options: SWACLIConfig) {
   }
 
   let projectConfig;
-  const detectedFolders = await detectProjectFolders(appLocation);
-  projectConfig = await generateConfiguration(detectedFolders.app[0], detectedFolders.api[0]);
   if (options.auto) {
     logger.log("Detecting build configuration...");
+    const detectedFolders = await detectProjectFolders(appLocation);
 
     if (detectedFolders.app.length === 0 && detectedFolders.api.length === 0) {
       logger.error(`Your app configuration could not be detected.`);
@@ -55,6 +54,7 @@ export async function build(options: SWACLIConfig) {
     }
 
     try {
+      projectConfig = await generateConfiguration(detectedFolders.app[0], detectedFolders.api[0]);
       appLocation = projectConfig.appLocation;
       apiLocation = projectConfig.apiLocation;
       outputLocation = projectConfig.outputLocation;
@@ -111,7 +111,7 @@ export async function build(options: SWACLIConfig) {
 
   collectTelemetryEvent(TELEMETRY_EVENTS.Build, {
     appFramework: projectConfig?.name?.split(", with")[0]!,
-    duration: (end - start).toLocaleString(),
+    duration: (end - start).toString(),
   });
 }
 
--- a/src/cli/commands/build/build.ts
+++ b/src/cli/commands/build/build.ts
@@ -10,11 +10,8 @@ import {
   runCommand,
   swaCliConfigFilename,
 } from "../../../core/utils";
-import { collectTelemetryEvent, getSessionId } from "../../../core/telemetry/utils";
-import { DEFAULT_CONFIG } from "../../../config";
-import os from "os";
-import { getMachineId } from "../../../core/swa-cli-persistence-plugin/impl/machine-identifier";
-import { TELEMETRY_BUILD_EVENT, TELEMETRY_MAC_ADDRESS_HASH_LENGTH } from "../../../core/constants";
+import { collectTelemetryEvent } from "../../../core/telemetry/utils";
+import { TELEMETRY_EVENTS } from "../../../core/constants";
 
 export async function build(options: SWACLIConfig) {
   const start = new Date().getTime();
@@ -112,12 +109,7 @@ export async function build(options: SWACLIConfig) {
   }
   const end = new Date().getTime();
 
-  collectTelemetryEvent(TELEMETRY_BUILD_EVENT, {
-    macAddressHash: (await getMachineId("sha256", TELEMETRY_MAC_ADDRESS_HASH_LENGTH)).toString(),
-    subscriptionId: DEFAULT_CONFIG.subscriptionId!,
-    sessionId: getSessionId(end),
-    OSType: os.type(),
-    OSVersion: os.version(),
+  collectTelemetryEvent(TELEMETRY_EVENTS.Build, {
     appFramework: projectConfig?.name?.split(", with")[0]!,
     duration: (end - start).toLocaleString(),
   });
--- a/src/cli/commands/build/build.ts
+++ b/src/cli/commands/build/build.ts
@@ -40,7 +40,6 @@ export async function build(options: SWACLIConfig) {
     return;
   }
 
-  let projectConfig;
   if (options.auto) {
     logger.log("Detecting build configuration...");
     const detectedFolders = await detectProjectFolders(appLocation);
@@ -53,6 +52,7 @@ export async function build(options: SWACLIConfig) {
       return showAutoErrorMessageAndExit();
     }
 
+    let projectConfig;
     try {
       projectConfig = await generateConfiguration(detectedFolders.app[0], detectedFolders.api[0]);
       appLocation = projectConfig.appLocation;
@@ -110,7 +110,6 @@ export async function build(options: SWACLIConfig) {
   const end = new Date().getTime();
 
   collectTelemetryEvent(TELEMETRY_EVENTS.Build, {
-    appFramework: projectConfig?.name?.split(", with")[0]!,
     duration: (end - start).toString(),
   });
 }
--- a/src/cli/commands/build/build.ts
+++ b/src/cli/commands/build/build.ts
@@ -10,8 +10,13 @@ import {
   runCommand,
   swaCliConfigFilename,
 } from "../../../core/utils";
+import { collectTelemetryEvent } from "../../../core/telemetry/utils";
+import { DEFAULT_CONFIG } from "../../../config";
+import { pkg } from "../..";
+import os from "os";
 
 export async function build(options: SWACLIConfig) {
+  const start = new Date().getTime();
   const workflowConfig = readWorkflowFile();
 
   logger.silly({
@@ -37,9 +42,11 @@ export async function build(options: SWACLIConfig) {
     return;
   }
 
+  let projectConfig;
+  const detectedFolders = await detectProjectFolders(appLocation);
+  projectConfig = await generateConfiguration(detectedFolders.app[0], detectedFolders.api[0]);
   if (options.auto) {
     logger.log("Detecting build configuration...");
-    const detectedFolders = await detectProjectFolders(appLocation);
 
     if (detectedFolders.app.length === 0 && detectedFolders.api.length === 0) {
       logger.error(`Your app configuration could not be detected.`);
@@ -49,9 +56,7 @@ export async function build(options: SWACLIConfig) {
       return showAutoErrorMessageAndExit();
     }
 
-    let projectConfig;
     try {
-      projectConfig = await generateConfiguration(detectedFolders.app[0], detectedFolders.api[0]);
       appLocation = projectConfig.appLocation;
       apiLocation = projectConfig.apiLocation;
       outputLocation = projectConfig.outputLocation;
@@ -104,6 +109,19 @@ export async function build(options: SWACLIConfig) {
     logger.log(`Building api with ${chalk.green(apiBuildCommand)} in ${chalk.dim(apiLocation)} ...`);
     runCommand(apiBuildCommand, apiLocation!);
   }
+  const end = new Date().getTime();
+  collectTelemetryEvent(
+    "build",
+    {
+      subscriptionId: DEFAULT_CONFIG.subscriptionId!,
+      CLIVersion: pkg.version,
+      OSType: os.platform(),
+      OSVersion: os.version(),
+      duration: (end - start).toLocaleString(),
+      appFramework: projectConfig?.name?.split(", with")[0]!,
+    },
+    { PID: process.pid }
+  );
 }
 
 function hasBuildOptionsDefined(options: SWACLIConfig): boolean {
--- a/src/cli/commands/build/build.ts
+++ b/src/cli/commands/build/build.ts
@@ -10,10 +10,11 @@ import {
   runCommand,
   swaCliConfigFilename,
 } from "../../../core/utils";
-import { collectTelemetryEvent } from "../../../core/telemetry/utils";
+import { collectTelemetryEvent, getSessionId } from "../../../core/telemetry/utils";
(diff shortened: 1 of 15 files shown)


rupareddy5-21:
Have to calculate sessionId from PID, will change in next iteration along with other fields.

src/cli/index.ts
```diff
@@ -19,7 +19,7 @@ import { registerTelemetry } from "./commands";
 
 export * from "./commands";
 
-const pkg = require("../../package.json");
+export const pkg = require("../../package.json");
```
sulabh-msft, line 22: why do we need to export the package.json?

rupareddy5-21: As we are using this 'pkg' in all the commands to collect CLI-Version field, instead of importing in all commands' files, I exported it from index.ts so it can be used wherever wanted.

src/core/constants.ts
```diff
@@ -181,6 +181,15 @@ export const SWA_CONFIG_FILENAME_LEGACY = "routes.json";
 export const CUSTOM_URL_SCHEME = "swa://";
 export const OVERRIDABLE_ERROR_CODES = [400, 401, 403, 404];
 
+export const TELEMETRY_MAC_ADDRESS_HASH_LENGTH = 32; //32 bytes
+export const MACHINE_ID_LENGTH = 16; //16 bytes
+
+export const TELEMETRY_LOGIN_EVENT = "login";
```
sgollapudi77, line 187: nit: Just a thought, can we have all these as a single const like this
export const TELEMETRY_EVENTS = {
  Login: "login",
  Start: "start"
}
and while using we can use like TELEMETRY_EVENTS.Login etc..,
Feel free to change/leave it as you see fit

sgollapudi77 (approved):

src/core/swa-cli-persistence-plugin/impl/machine-identifier.ts
```diff
@@ -22,10 +22,10 @@ export async function getMachineId(): Promise<string> {
  * Get the mac address of the machine and hash it.
  * @returns {Promise<string>} A 32-byte hash of the mac address.
  */
```
sulabh-msft, line 24: 32 bytes.. is the jsdoc correct ?Shall we update it ?

src/core/swa-cli-persistence-plugin/impl/machine-identifier.ts
```diff
@@ -22,10 +22,10 @@ export async function getMachineId(): Promise<string> {
...
-async function getMacMachineId(): Promise<string | undefined> {
+async function getMacMachineId(algorithm: string, length: number): Promise<string | undefined> {
```
sulabh-msft, line 25: do we support all algorithms are we need to restrict it to some fixed set ?

src/core/swa-cli-persistence-plugin/persistence-cache-plugin.ts
```diff
@@ -30,7 +31,7 @@ export class SWACLIPersistenceCachePlugin implements ICachePlugin {
   public async beforeCacheAccess(cacheContext: TokenCacheContext): Promise<void> {
     logger.silly(`Executing before cache access plugin`);
 
-    const machineId = await getMachineId();
+    const machineId = await getMachineId("shake256", MACHINE_ID_LENGTH);
```
sulabh-msft, line 34: any reasons for using different algo for machineId here and what we are logging ? should we use same machineId ?

src/cli/commands/start/start.ts
```diff
@@ -298,4 +306,20 @@ export async function start(options: SWACLIConfig) {
     .catch((err) => {
       logger.error(err.message, true);
     });
+
+  const end = new Date().getTime();
+
+  collectTelemetryEvent(
+    TELEMETRY_START_EVENT,
+    {
+      macAddressHash: (await getMachineId("sha256", TELEMETRY_MAC_ADDRESS_HASH_LENGTH)).toString(),
```
sulabh-msft, line 315: Can these properties be added at the common layer so that we don't need to force every caller of `collectTelemetryEvent` to send this information if it is common to all the telemetry events

src/cli/commands/start/start.ts
```diff
@@ -298,4 +306,20 @@ export async function start(options: SWACLIConfig) {
...
+      subscriptionId: DEFAULT_CONFIG.subscriptionId!,
+      sessionId: getSessionId(end),
+      OSType: os.type(),
+      OSVersion: os.version(),
+      apiRuntime: swaConfigFileContent?.platform.apiRuntime!,
+      duration: (end - start).toLocaleString(),
```
sulabh-msft, line 321: toLocalString() ? or should it be locale agnostic ?

src/core/constants.ts
```diff
@@ -181,6 +181,15 @@ export const SWA_CONFIG_FILENAME_LEGACY = "routes.json";
 export const CUSTOM_URL_SCHEME = "swa://";
 export const OVERRIDABLE_ERROR_CODES = [400, 401, 403, 404];
 
+export const TELEMETRY_MAC_ADDRESS_HASH_LENGTH = 32; //32 bytes
+export const MACHINE_ID_LENGTH = 16; //16 bytes
```
sulabh-msft, line 185: Why do we need two different hash length for the MAC

src/cli/commands/start/start.ts
sulabh-msft, line 315: Also see if we can get the MAC Id once for a session and use for all the telemetry events

src/cli/commands/init/init.ts
```diff
@@ -130,6 +136,17 @@ export async function init(options: SWACLIConfig, showHints: boolean = true) {
 
     logger.log(`- Use ${chalk.cyan("swa deploy")} to deploy your app to Azure.\n`);
   }
+  const end = new Date().getTime();
+
+  await collectTelemetryEvent(TELEMETRY_INIT_EVENT, {
+    macAddressHash: (await getMachineId("sha256", TELEMETRY_MAC_ADDRESS_HASH_LENGTH)).toString(),
+    subscriptionId: DEFAULT_CONFIG.subscriptionId!,
+    sessionId: getSessionId(end),
```
sulabh-msft, line 144: looks little or for us to take end time as the sessionId.. Any reason for that.. Also can it be calculated in the telemetry util or during the initialization of the telemetry library. Session is a uniqueid for the cli invocation?

sulabh-msft (changes requested):
Reviewed the PR. Few changes requested.

src/core/swa-cli-persistence-plugin/impl/machine-identifier.ts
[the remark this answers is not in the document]
rupareddy5-21: our purpose of logging MacAddressHash is to join the VScode telemetry table which needs a SHA256 hash in 32bytes and the existing getMacMachineId uses a different algo and different hashlength.

src/cli/commands/init/init.ts
rupareddy5-21: Yes, SessionId is the unique Id for each time CLI is invoked. we are calculating it as a combination of ProcessId and timeStamp and get the hash of the combined string. Used `end` as the timestamp as I wanted to calculate the session Id at the time when the call to collecttelemetryEvent is made.

src/core/telemetry/utils.ts
```diff
@@ -5,13 +5,31 @@ import TelemetryReporter from "./telemetryReporter";
...
-    reporter.sendTelemetryEvent(event, properties, measurements);
+    let macAddressHash = (await getMachineIdForTelemetry()).toString();
+    let extendedTelemetryEventProperties = {
+      macAddressHash: macAddressHash,
+      installationId: crypto
+        .createHash("sha256")
+        .update(pkg.version + macAddressHash)
+        .digest("hex"),
+      subscriptionId: DEFAULT_CONFIG.subscriptionId!,
+      sessionId: getSessionId(new Date().getTime()),
```
sulabh-msft, line 27: Q: Should we have same sessionId if two telemetry ponts are captured in the same session.. This will not have same session. ?

src/core/telemetry/utils.ts
```diff
@@ -5,13 +5,31 @@ import TelemetryReporter from "./telemetryReporter";
...
 
 const aiKey = "8428a7f6-6650-4490-a15a-c7f7a16449d7";
+const pkg = require("../../../package.json");
 
 export async function collectTelemetryEvent(event: string, properties?: TelemetryEventProperties, measurements?: TelemetryEventMeasurements) {
   const reporter = await getTelemetryReporter();
+
   if (reporter) {
-    reporter.sendTelemetryEvent(event, properties, measurements);
+    let macAddressHash = (await getMachineIdForTelemetry()).toString();
```
sulabh-msft, line 19: `const`

src/cli/commands/build/build.ts
```diff
@@ -37,9 +40,11 @@ export async function build(options: SWACLIConfig) {
     return;
   }
 
+  let projectConfig;
+  const detectedFolders = await detectProjectFolders(appLocation);
+  projectConfig = await generateConfiguration(detectedFolders.app[0], detectedFolders.api[0]);
```
sulabh-msft, line 45: `  projectConfig = await generateConfiguration(detectedFolders.app[0], detectedFolders.api[0]);
`
why is it no longer needed to be part of try catch ?

src/cli/commands/build/build.ts
```diff
@@ -104,6 +107,12 @@ export async function build(options: SWACLIConfig) {
     logger.log(`Building api with ${chalk.green(apiBuildCommand)} in ${chalk.dim(apiLocation)} ...`);
     runCommand(apiBuildCommand, apiLocation!);
   }
+  const end = new Date().getTime();
+
+  collectTelemetryEvent(TELEMETRY_EVENTS.Build, {
+    appFramework: projectConfig?.name?.split(", with")[0]!,
+    duration: (end - start).toLocaleString(),
```
sulabh-msft, line 114: Why are we converting to LocaleString.. it will be different for different locales right.. so do we want different duration format in our logs ?

src/cli/commands/deploy/deploy.ts
```diff
@@ -291,6 +297,16 @@ export async function deploy(options: SWACLIConfig) {
...
   } finally {
+    const end = new Date().getTime();
+
+    collectTelemetryEvent(
+      TELEMETRY_EVENTS.Deploy,
+      {
+        apiRuntime: swaConfigFileContent?.platform?.apiRuntime!,
+        duration: (end - start).toLocaleString(),
+      },
+      { appRuntime: nodeMajorVersion }
```
sulabh-msft, line 308: Can we put the appRuntime to as part  apiRuntime properties bag

src/core/telemetry/utils.ts
```diff
@@ -4,28 +4,56 @@ dotenv.config();
...
+    let extendedTelemetryEventProperties = {
+      macAddressHash: macAddressHash,
+      installationId: crypto
+        .createHash("sha256")
+        .update(pkg.version + macAddressHash)
+        .digest("hex"),
+      subscriptionId: DEFAULT_CONFIG.subscriptionId!,
+      sessionId: getSessionId(new Date().getTime()),
+      OSType: os.type(),
```
sulabh-msft, line 28: osType, osVersion ?

src/core/telemetry/utils.ts
```diff
@@ -4,28 +4,56 @@ dotenv.config();
...
+      OSVersion: os.version(),
+    } as TelemetryEventProperties;
+
+    reporter.sendTelemetryEvent(event, { ...extendedTelemetryEventProperties, ...properties }, measurements);
```
sulabh-msft, line 32: Q: what will happen if some one passes ostype/OSVersion etc will it be override what we are calculating here?

src/cli/commands/start/start.ts
```diff
@@ -298,4 +303,15 @@ export async function start(options: SWACLIConfig) {
...
+
+  const end = new Date().getTime();
+
+  collectTelemetryEvent(
+    TELEMETRY_EVENTS.Start,
+    {
+      apiRuntime: swaConfigFileContent?.platform.apiRuntime!,
+      duration: (end - start).toLocaleString(),
+    },
+    { appRuntime: nodeMajorVersion }
```
sulabh-msft, line 315: how the metric will look like for tappRuntime ?

rupareddy5-21: It'll look like node 16

src/core/telemetry/utils.ts
rupareddy5-21: users won't have control over what fields we log for telemetry, so they cant pass OsType or version, we'll only log what we are calculating

src/core/telemetry/utils.ts
rupareddy5-21: but in any possible case flow, we'll only be collecting one telemetry point right?

src/core/swa-cli-persistence-plugin/impl/machine-identifier.ts
rupareddy5-21: Wassim mentioned he'll be updating the comments

src/core/telemetry/utils.ts
sulabh-msft: It might be possible to collect one or more telemetry point for a flow. e.g. we starte perf measures and then these logs or multiple log points for the same flow.. in that case the sessionId will be different.

 So in that case we should have same sessionId for both telemetry points..

src/core/telemetry/utils.ts
sulabh-msft: My question was do we have a way to guard against such unitentional action on the caller side. How about 

`{  ...properties, ...extendedTelemetryEventProperties },`

src/cli/commands/build/build.ts
```diff
@@ -104,6 +107,12 @@ export async function build(options: SWACLIConfig) {
     logger.log(`Building api with ${chalk.green(apiBuildCommand)} in ${chalk.dim(apiLocation)} ...`);
     runCommand(apiBuildCommand, apiLocation!);
   }
+  const end = new Date().getTime();
+
+  collectTelemetryEvent(TELEMETRY_EVENTS.Build, {
+    appFramework: projectConfig?.name?.split(", with")[0]!,
```
sulabh-msft, line 113: can we ensure that this code doesn't throw null reference

src/core/telemetry/utils.ts
rupareddy5-21: yes my bad, got your point, will update

src/cli/commands/init/init.ts
```diff
@@ -130,6 +133,18 @@ export async function init(options: SWACLIConfig, showHints: boolean = true) {
 
     logger.log(`- Use ${chalk.cyan("swa deploy")} to deploy your app to Azure.\n`);
   }
+  const end = new Date().getTime();
+  const appFramework = projectConfig?.name?.split(", with")[0]!;
+  if (appFramework) {
+    await collectTelemetryEvent(TELEMETRY_EVENTS.Init, {
+      appFramework: appFramework,
```
sulabh-msft, line 140: why do we need this if-else logic ? why cannot we have appFramework logged with undefined/null value as we are doing for api runtime property at other places?

src/cli/commands/start/start.ts
```diff
@@ -25,6 +28,7 @@ export async function start(options: SWACLIConfig) {
   // Code below doesn't have access to these environment variables which are defined later below.
   // Make sure this code (or code from utils) does't depend on environment variables!
 
+  const start = new Date().getTime();
```
sulabh-msft, line 31: at all the places.. shall we name the variable start/end  as cmdStartTime/ cmdEndTime to be more clear on what these start/end refer to?

src/core/constants.ts
```diff
@@ -181,6 +181,17 @@ export const SWA_CONFIG_FILENAME_LEGACY = "routes.json";
...
```
sulabh-msft, line 185: Add a comment /warning to not change these values

src/core/telemetry/utils.ts
```diff
@@ -4,28 +4,55 @@ dotenv.config();
...
-    reporter.sendTelemetryEvent(event, properties, measurements);
+    const macAddressHash = (await getMachineIdForTelemetry()).toString();
+    let extendedTelemetryEventProperties = {
+      macAddressHash: macAddressHash,
+      installationId: crypto
+        .createHash("sha256")
+        .update(pkg.version + macAddressHash)
+        .digest("hex"),
+      subscriptionId: DEFAULT_CONFIG.subscriptionId!,
+      OsType: os.type(),
```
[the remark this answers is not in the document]
sulabh-msft: subscriptionId can change as user can select different subscriptionId.. so its ok to keep it here.

sulabh-msft (approved):
Approved the PR with suggestions. Please resolve them before merging the PR

src/cli/commands/build/build.ts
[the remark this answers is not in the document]
rupareddy5-21: will address in next iteration

rupareddy5-21: will address in next iteration

Issue #663, opened 2023-03-03
[start] Start command does not connect to web or api server on Node 18


sinedied:
Before filing this issue, please ensure you're using the latest CLI by running `swa --version` and comparing to the latest version on [npm](https://www.npmjs.com/package/@azure/static-web-apps-cli).

**Are you accessing the CLI from the default port `:4280` ?**

- [ ] No, I am using a different port number (`--port`) and accessing the CLI from that port
- [x] Yes, I am accessing the CLI from port `:4280`

> Make sure you are accessing the URL printed in the console when running `swa start`!

**ℹ️ NOTE: Make sure to enable debug logs when running any `swa` commands using `--verbose=silly`**

**Describe the bug**
On Node 18, `npm start` never connects to a web or api server, it's stuck until timeout. On Node 16, no issue.

This is related to this issue https://github.com/jeffbski/wait-on/issues/137 of the `wait-on` package.

sgollapudi77:
Hi @sinedied , is there any alternative that we can use instead of "wait-on" package?

sinedied:
> Hi @sinedied , is there any alternative that we can use instead of "wait-on" package?

None that I know of. This is problematic in the whole Node.js ecosystem as they made the changed the priority order from ipv4 to ipv6.
We could work around that on our side but it would be a bit convoluted (trying both localhost and 127.0.0.1, and connecting to the first one that succeed.

> As a temporary workaround you can use 127.0.0.1 in stead of localhost for the API and/or for the frontend, use it like this:
> 
> > swa start --host=127.0.0.1 --app-devserver-url=http://127.0.0.1:3010
> 
> Mind that the port number 3010 can be different on your side of course, and you should add your own extra parameters.

You're right, but the `--host` isn't needed, only `--app-devserver-url=http://127.0.0.1:<PORT>` or `--api-devserver-url=http://127.0.0.1:<PORT>`.

One issue is that once the `--api-devserver-url` option is set, the Function API emulator doesn't start automatically anymore and need to be run manually.

rodyvansambeek:
> One issue is that once the `--api-devserver-url` option is set, the Function API emulator doesn't start automatically anymore and need to be run manually.

You are right, but that's why you include the --host in combination with the --api-location option (and not the --api-devserver-url option), in my example I omitted the api-location option, but it should look something like: 

[...]

PR #704, opened 2023-05-31
fix: waitOn and fetch IPv4&v6 instead of localhost

cjk7989:
Fix: issue [#663](https://github.com/Azure/static-web-apps-cli/issues/663).
Use 127.0.0.1 and [::1] for waitOn() and fetch() when using "swa start src --api-location api" with the default --host=localhost. Then check if any of them successes. This is because localhost may not be resolved to ipv4 on some systems.

### DIFF
--- a/src/core/utils/net.ts
+++ b/src/core/utils/net.ts
@@ -104,32 +104,35 @@ export async function validateDevServerConfig(url: string | undefined, timeout: 
     if (resolvedPortNumber !== 0) {
       const spinner = ora();
       const waitOnOneOfResources = hostname === "localhost" ? [`tcp:127.0.0.1:${port}`, `tcp:[::1]:${port}`] : [`tcp:${hostname}:${port}`];
-      let isConnected = false;
       spinner.start(`Waiting for ${chalk.green(url)} to be ready`);
-      for (let resource of waitOnOneOfResources) {
-        try {
-          await waitOn({
-            resources: [resource],
-            delay: 1000, // initial delay in ms, default 0
-            interval: 100, // poll interval in ms, default 250ms
-            simultaneous: 1, // limit to 1 connection per resource at a time
-            timeout: timeout ? timeout * 1000 : timeout, // timeout in ms, default Infinity
-            tcpTimeout: 1000, // tcp timeout in ms, default 300ms
-            window: 1000, // stabilization time in ms, default 750ms
-            strictSSL: false,
-            verbose: true, // force disable verbose logs even if SWA_CLI_DEBUG is enabled
+
+      let promises = waitOnOneOfResources.map((resource) => {
+        return waitOn({
+          resources: [resource],
+          delay: 1000, // initial delay in ms, default 0
+          interval: 100, // poll interval in ms, default 250ms
+          simultaneous: 1, // limit to 1 connection per resource at a time
+          timeout: timeout ? timeout * 1000 : timeout, // timeout in ms, default Infinity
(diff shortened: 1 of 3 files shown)


src/core/utils/net.ts
```diff
@@ -103,22 +103,33 @@ export async function validateDevServerConfig(url: string | undefined, timeout:
...
-          tcpTimeout: 1000, // tcp timeout in ms, default 300ms
-          window: 1000, // stabilization time in ms, default 750ms
-          strictSSL: false,
-          verbose: false, // force disable verbose logs even if SWA_CLI_DEBUG is enabled
-        });
-        spinner.succeed(`Connected to ${chalk.green(url)} successfully`);
+      const waitOnOneOfResources = hostname === "localhost" ? [`tcp:127.0.0.1:${port}`, `tcp:[::1]:${port}`] : [`tcp:${hostname}:${port}`];
+      let isConnected = false;
+      spinner.start(`Waiting for ${chalk.green(url)} to be ready`);
+      for (let resource of waitOnOneOfResources) {
```
sinedied, line 109: Using `for` with `await` is an anti-pattern as it will wait for the first iteration to be resolved before moving on to the second one, not running both promises in parallel (as we want to).

I suggest to refactor it something like this instead:
```js
const promises = waitOnOneOfResources.map(
  resource => {
    return waitOn(...)
      .then(() => resource)
      .catch(err => {
        logger.silly(`Could not connect to ${resource}`);
        throw err;
      });
  }
);
const resource = await Promise.any(promises);
if (resource instanceof AggregateError) {
  // both promises rejected, it failed
} else {
  // resource is connected
}
```

src/msha/handlers/function.handler.ts
```diff
@@ -53,8 +54,29 @@ function injectClientPrincipalCookies(req: http.ClientRequest) {
   }
 }
 
+async function resolveLocalhostByFetching(url: string) {
+  const { protocol, port } = parseUrl(url);
+  const fetchOneOfUrl = [`${protocol}//127.0.0.1:${port}`, `${protocol}//[::1]:${port}`];
+
+  for (let Url of fetchOneOfUrl) {
```
sinedied, line 61: Again, be careful with using `await` in `for` loops as it breaks asynchronicity.
You can use the same pattern as above to keep it asynchronous.

sinedied (changes requested):
There's an issue using `await` in a `for` loop, but looks good otherwise

src/core/utils/net.ts
cjk7989: The await mthod is modified in the second commit "correct the await method". Thanks for your suggestion.

src/msha/handlers/function.handler.ts
cjk7989: Done. Thanks!

src/msha/handlers/function.handler.ts
```diff
@@ -53,8 +54,40 @@ function injectClientPrincipalCookies(req: http.ClientRequest) {
...
+        }
+      })
+      .catch((err) => {
+        logger.silly(`Could not fetch ${Url}`);
+        throw err;
+      });
+  });
+
+  try {
+    const avaliableUrl = await Promise.any(promises);
```
sinedied, line 79:
```suggestion
    const availableUrl = await Promise.any(promises);
```
Small typo

sinedied (approved):
Sent suggestion for small typo, otherwise LGTM 👍

PR #729, opened 2023-07-18
fix: fix the logic of installing cli in dockerfile

cjk7989:


cjk7989:
@all-contributors add @cjk7989 code

Issue #733, opened 2023-07-19
apiLocation Cannot Resolve localhost


PeytonRG:
Before filing this issue, please ensure you're using the latest CLI by running `swa --version` and comparing to the latest version on [npm](https://www.npmjs.com/package/@azure/static-web-apps-cli).

**Are you accessing the CLI from the default port `:4280` ?**

- [ ] No, I am using a different port number (`--port`) and accessing the CLI from that port
- [x] Yes, I am accessing the CLI from port `:4280`

> Make sure you are accessing the URL printed in the console when running `swa start`!

**ℹ️ NOTE: Make sure to enable debug logs when running any `swa` commands using `--verbose=silly`**

**Describe the bug**
After updating to versions of the SWA CLI greater than 1.1.2, my CLI fails to start up when the apiLocation is set to `localhost`. I first encountered this when updating to 1.1.3 on Node 16.20.1, and it still happens on 1.1.4. I have tested this on both Node 16.20.1 and Node 18.17.0 with the same results, since 1.1.4 includes a fix for Node 18. If I change appDevserverUrl and apiLocation from `localhost` to `127.0.0.1`, the CLI will start up as normal, but this was not required on 1.1.2 and earlier.

**To Reproduce**

1. Set `swa-cli.config.json` appDevserverUrl and apiLocation to `localhost`, like so:
``` JSON
{
  "$schema": "https://aka.ms/azure/static-web-apps-cli/schema",
  "configurations": {
    "web-app": {
      "appLocation": ".",
      "outputLocation": "build",
      "appBuildCommand": "npm run build",
      "run": "npm start",
      "appDevserverUrl": "http://localhost:3000",
      "apiLocation": "http://localhost:5003"
    }
  }
}

```
2. Type in command 'swa start'
3. See error

**Expected behavior**
The SWA CLI should start up normally.

**Screenshots**

**Desktop (please complete the following information):**

- OS: macOS
- Version 13.4.1 (c)

**Additional context**
As I mentioned above, a `swa-cli.config.json` with 127.0.0.1 rather than localhost allows the cli to start normally, as it did on 1.1.2 and earlier versions
``` JSON
{
  "$schema": "https://aka.ms/azure/static-web-apps-cli/schema",
  "configurations": {
    "web-app": {
      "appLocation": ".",
      "outputLocation": "build",
      "appBuildCommand": "npm run build",
      "run": "npm start",
      "appDevserverUrl": "http://localhost:3000",
[...]

PR #737, opened 2023-08-02
fix: accept apiDevserverUrl with non-ok response

cjk7989:
This PR fixs [ISSUE #733](https://github.com/Azure/static-web-apps-cli/issues/733). If an api dev server are used, and the apiDevserverUrl can be fetched, swa-cli will accept it even if the response is not ok, for example 404 not found.


v1212 (approved):

itpropro:
Hey @cjk7989, any timeline on a release of this fix?

PR #740, opened 2023-08-18
Jikun/start or deploy from outside

cjk7989:
Supported "swa start app1" or "swa deploy app1" outside the app1 folder.
For example, under the current working dir containing two projects "app1" and "app2", the above commands will work if there is a "swa-cli.config.json" in "app1" folder, while in the past users must run "swa start" under the app folder.

### DIFF
--- a/src/core/utils/cli-config.ts
+++ b/src/core/utils/cli-config.ts
@@ -49,20 +49,27 @@ export function matchLoadedConfigName(name: string) {
 export async function getConfigFileOptions(configName: string | undefined, configFilePath: string): Promise<SWACLIConfig> {
   logger.silly(`Getting config file options from ${configFilePath}...`);
 
-  configFilePath = path.resolve(configFilePath);
-  if (!swaCliConfigFileExists(configFilePath)) {
+  let resolvedConfigFilePath = path.resolve(configFilePath);
+  if (!swaCliConfigFileExists(resolvedConfigFilePath)) {
     logger.silly(`Config file does not exist at ${configFilePath}`);
-    return {};
+    if (configName) {
+      resolvedConfigFilePath = path.resolve(configName, configFilePath);
+      if (!swaCliConfigFileExists(resolvedConfigFilePath)) {
+        return {};
+      } else {
+        logger.silly(`Found config file at ${resolvedConfigFilePath}`);
+      }
+    }
   }
 
-  const cliConfig = await tryParseSwaCliConfig(configFilePath);
+  const cliConfig = await tryParseSwaCliConfig(resolvedConfigFilePath);
   if (!cliConfig.configurations) {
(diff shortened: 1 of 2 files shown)


src/core/utils/cli-config.ts
```diff
@@ -47,22 +48,37 @@ export function matchLoadedConfigName(name: string) {
...
-    return {};
+  let resolvedConfigFilePath = path.resolve(configFilePath);
+  if (!swaCliConfigFileExists(resolvedConfigFilePath)) {
+    logger.silly(`Config file does not exist at "${configFilePath}"`);
+
+    // Handle the case when the user runs the command outside the project path
+    if (!configName || !swaCliConfigFileExists(path.resolve(configName, configFilePath))) {
+      return {};
+    }
+    logger.warn(`WARNING: Config file does not exist at "${configFilePath}", but can be detected at "${path.join(configName, configFilePath)}".`);
```
IvanJobs, line 61: It's a little strange that both `path.join` and `path.resolve` were used in the same part of code.  I believe using only one of them is enough depending on the case.

Nit: make it as a variable in the beginning.

src/core/utils/cli-config.ts
```diff
@@ -47,22 +48,37 @@ export function matchLoadedConfigName(name: string) {
...
+    const { confirmConfigPath } = await promptOrUseDefault(false, {
+      type: "confirm",
+      name: "confirmConfigPath",
+      message: `Do you mean "swa --config ${path.join(configName, configFilePath)} <command> ${configName} [options]"?`,
+      initial: true,
+    });
+    if (!confirmConfigPath) {
+      return {};
+    }
+    resolvedConfigFilePath = path.resolve(configName, configFilePath);
```
IvanJobs, line 71: Same question: Why in some places `path.resolve` was used and in some other places `path.join` was used?

src/core/utils/cli-config.ts
```diff
@@ -76,10 +92,10 @@ export async function getConfigFileOptions(configName: string | undefined, confi
     }
 
     const [configName, config] = Object.entries(cliConfig.configurations)[0];
-    printConfigMsg(configName, configFilePath);
+    printConfigMsg(configName, resolvedConfigFilePath);
     currentSwaCliConfigFromFile = {
       name: configName,
-      filePath: configFilePath,
+      filePath: resolvedConfigFilePath,
```
IvanJobs, line 98: Suggestion:

You can handle the config path fallback logic in the beginning of this function, so that you won't need to change `configFilePath` to `resolved*` all the way down here.

IvanJobs:
Just checked the code in details. It seems we got a config path option in SWA CLI's global configs. If Cx runs SWA outside of a project folder, we expect Cx to specify the config path as an extra option. This may be the appropriate way to handle this from a PM's perspective.

IvanJobs (approved):

PR #758, opened 2023-10-24
fix: revert changes that cause slow proxying

johnnyreilly:
This PR addresses https://github.com/Azure/static-web-apps-cli/issues/736 which concerns slow proxied requests for static content.

These slow proxied requests were caused by the changes made by @cjk7989 in https://github.com/Azure/static-web-apps-cli/pull/720

Whilst well intended, they made the SWA CLI very difficult to use for local development as of 1.1.4 when the changes were released.

Looking at the comments in https://github.com/Azure/static-web-apps-cli/issues/736 it's clear this is affecting many people.  This PR reverts that change.

Speaking for my own organisation, we have widely pinned to 1.1.3.

## Motivations

1. The implemented change has not resulted an improved development experience
2. The origin motivation behind the changes in https://github.com/Azure/static-web-apps-cli/pull/720 was to resolve https://github.com/Azure/static-web-apps-cli/issues/663.  Those issues arose with early versions of Node.js 18 that were affecting `wait-on`.  That issue is now resolved with Node.js according this this: https://github.com/jeffbski/wait-on/issues/137#issuecomment-1775289798 - Node.js 20 is now the LTS version of Node.js.

PR #761, opened 2023-10-31
fix: remove initial delay to speed up requests

cjk7989:
The previous wait-on function has a param: initial delay = 1000, which caused the waiting time of a request to a dev server to at least 1000ms. This PR remove the param, which will be 0 by default.


johnnyreilly:
Testing results can be seen here: https://github.com/Azure/static-web-apps-cli/pull/758#issuecomment-1787950192

Issue #715, opened 2023-06-16
Utilize az cli/managed identity credentials


onionhammer:
swa login should be able to use the same credentials available to a managed identity (running on something like app service or azure container app), or azure CLI when using the SWA CLI locally 

"@azure/identity" is used here somewhere, but it seems to do this instead (from within a docker container)

swa login -u

```
Checking Azure session...
✖ Failed to setup project: Cannot autolaunch D-Bus without X11 $DISPLAY

✖ If you believe this behavior is unexpected, please raise a GitHub issue at:
  https://github.com/Azure/static-web-apps-cli/issues/new/choose
```

adrianhall:
#774 is merged and corrects this issue as well.

PR #774, opened 2023-11-04
feat: Support Azure CLI credential

### LINKED ISSUE #751: Support reusing credentials from azure/login action

I'm trying to use the SWA CLI in order to have more control over the the build and deployment of multiple environments in GitHub. So I'm not using the GitHub action. I normally use the federated login from Github in order to perform tasks in workflows and the azure/login action. Doing this, I don't need to setup client secrets. But this fails:

```yaml
... Build steps ...
- name: Azure login
  uses: azure/login@v1
  with:
    client-id: ${{ vars.APP_CLIENT_ID }}
    tenant-id: ${{ vars.APP_TENANT_ID }}
    allow-no-subscriptions: true
- name: Deploy
  run: |
    npm i -g @azure/static-web-apps-cli
    swa deploy --no-use-keychain --config-name ${{matrix.environment}} --env Production
```

It gets stuck trying to open a browser and wait for a device code login. SWA CLI only accepts client secrets as an identity.

I wanted the SWA CLI to reuse the credentials already set in the azure/login action so it doesn't need to use client secrets.

So far only alternative today is to use client secrets

**Additional context**

shibayan:
The current static-web-apps-deploy task does not support Service Principal, so we always need to use tokens for deployment, but it is very time consuming to configure everything when deploying to multiple SWAs.

If the SWA CLI supports Azure CLI login (azure/login action), deployment would be very simple.

Close #751

### DIFF
--- a/src/core/account.ts
+++ b/src/core/account.ts
@@ -64,7 +64,7 @@ export async function authenticateWithAzureIdentity(details: LoginDetails = {}, 
   const environmentCredential = new EnvironmentCredential();
 
   const azureCliCredential = new AzureCliCredential({
-    tenantId: details.clientId,
+    tenantId: details.tenantId,
   });
 
   // Only use interactive browser credential if we're not running in docker
--- a/src/core/account.ts
+++ b/src/core/account.ts
@@ -2,6 +2,7 @@ import { StaticSiteARMResource, WebSiteManagementClient } from "@azure/arm-appse
 import { GenericResourceExpanded, ResourceGroup, ResourceManagementClient } from "@azure/arm-resources";
 import { Subscription, SubscriptionClient } from "@azure/arm-subscriptions";
 import {
+  AzureCliCredential,
   ChainedTokenCredential,
   ClientSecretCredential,
   DeviceCodeCredential,
@@ -62,8 +63,12 @@ export async function authenticateWithAzureIdentity(details: LoginDetails = {}, 
 
   const environmentCredential = new EnvironmentCredential();
 
+  const azureCliCredential = new AzureCliCredential();
+
   // Only use interactive browser credential if we're not running in docker
-  const credentials = isRunningInDocker() ? [environmentCredential, deviceCredential] : [environmentCredential, browserCredential, deviceCredential];
+  const credentials = isRunningInDocker()
+    ? [azureCliCredential, environmentCredential, deviceCredential]
+    : [azureCliCredential, environmentCredential, browserCredential, deviceCredential];
 
   if (details.tenantId && details.clientId && details.clientSecret) {
     const clientSecretCredential = new ClientSecretCredential(details.tenantId, details.clientId, details.clientSecret, {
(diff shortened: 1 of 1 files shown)


rhuanbarreto:
@manekinekko @v1212 @cjk7989 @IvanJobs any updates? This is important for improvements in static web apps deployment

shibayan:
@manekinekko @v1212 @cjk7989 @IvanJobs I would be happy to have this feature reviewed as I believe it is something that many, including myself, have been waiting for.

cjk7989:
Thanks @shibayan, sorry for the delay. I'll test the changes from my side and publish in the next release (maybe version 1.1.8). Could you give some details about the scenario how you login swa using Azure CLI?

shibayan:
@cjk7989 Thanks, I believe the biggest scenario where we want to use Azure CLI credentials from SWA CLI is OpenID Connect (Federated Credential) in GitHub Actions as per #751 #715.

Web App deployments use Federated Credentials via Azure CLI, but since they are not available for Static Web Apps deployments, it is necessary to manage deployment tokens separately.

rhuanbarreto:
Just to raise my case. I'm using the CLI directly because I have a monorepo deploying many different static websites. So I already use the azure login action in github actions to issue a token against azure management.

cjk7989:
Hi @shibayan,  sorry for the delay and thanks for the great contribution! I plan to test it from my side and apply the feature in the next release. Could you share a complete scenario how do you set up azure login in GitHub Actions?

shibayan:
@cjk7989 In a monolipo configuration, it is not reasonable to manage credentials such as deployment tokens separately, since a single repository contains multiple applications and deployment is required for each.

The following is a workflow for a simple application I am running, which uses Federated Credential authentication in combination with Azure Login for the Web App, so only one set of credentials is needed, making the configuration even more secure.

- https://github.com/shibayan/appserviceinfo/blob/master/.github/workflows/backend-deploy.yml

The following is a deployment to Static Web App. The standard Action is provided as a Docker Image that includes a build environment, so the overhead is large for configurations where you want to build in advance and leave deployment to the user, so deployment is done using the SWA CLI.

- https://github.com/shibayan/appserviceinfo/blob/master/.github/workflows/frontend-deploy.yml

Since deployments to the same resource group will be made, we believe it is a common requirement that deployments to Static Web App use the same authentication information as Web App to reduce administration.

Although not a public project, the business is developing four SWAs in monolipo and deployments require four deployment tokens. It is extremely difficult to manage these four without making mistakes.

KoblerS:
Any news on this PR, I would really appreciate if this feature can be merged as I really need to use the local Azure credentials?

rhuanbarreto:
@manekinekko @v1212 @cjk7989 @IvanJobs we really need this in order to simplify our CI workflows. How can we find a way to prioritize this?

shibayan:
Any updates?

Timothyw0:
Hi All, sorry for the delay in review. Thank you @shibayan for raising this PR and adding support for the Azure/login Github action. 

We will include this in our next release :)

PR #786, opened 2023-12-05
Refreshing Node.js supported versions

### LINKED ISSUE #756: Node 20 support
If the staticwebapp.config.json file contains 

```
  "platform": {
    "apiRuntime": "node:20"
  },
```

The following error message will appear:

```
[swa] WARNING: Failed to read staticwebapp.config.json configuration from:
[swa]    /YOURAPP/staticwebapp.config.json
[swa]
[swa] The following error was encountered: must be equal to one of the allowed values dotnet:3.1, dotnet:6.0, dotnet-isolated:6.0, dotnet-isolated:7.0, node:12, node:14, node:16, node:18, python:3.8, python:3.9, python:3.10
[swa] 0: {
[swa] 1:     "platform": {
[swa] 2:         "apiRuntime": "node:20"
[swa] 3:     },
[swa]
[swa] Please fix the above error and try again to load and use the configuration.
[swa] Read more: https://aka.ms/swa/config-schema
```

Node 20 is the current release since march and will be LTS in 4 days, so swa should support it.

Update: Node 20 is now the LTS version of Node, 18 is only receiving security fixes and 16 is EOL.

aaronpowell:
Updating the supported node versions to match the [current list](https://learn.microsoft.com/en-gb/azure/azure-functions/functions-reference-node?tabs=javascript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#supported-versions), including previews.

Fixes #756


aaronpowell:
Oh right, tests probably need updating too 😅

aaronpowell:
I'm not sure how to fix the failing test. The test that fails is [this one here](https://github.com/Azure/static-web-apps-cli/blob/main/src/core/func-core-tools.spec.ts#L230-L247) and it fails because it "thinks" we have v3 of the core tools installed but we need v4 (since v3 didn't support Node18 but now v3 does).

The problem is that we now have some overlaps in the version support (v3 and v4 both support Node 18 and Node 20), resulting in it not being able to tell what to install.

Probably need an overhaul of that test but I'm not sure what it should look like.

adrianhall:
Hold off on merging this one for right now

Hey @aaronpowell - I'll get this merged once we have the big CVE/ESM move done.

Timothyw0:
@aaronpowell Could you please change the cached version here that will fix the unit test? In src/core/func-core-tools.spec.ts line 227:
`fct.setCachedInstalledSystemCoreToolsVersion(2);`

This will force the binary to be out of sync with the version of Node and return the expected downloaded binary

Timothyw0:
@aaronpowell Based on the timeline of our upcoming release, I've created a PR including your changes and the unit test fix at #869

Issue #754, opened 2023-10-11
Failed to load staticwebapp.config.json schema on start


StanSymphonia:
I'm using the latest version 1.1.4

**Are you accessing the CLI from the default port `:4280` ?**

- [ ] No, I am using a different port number (`--port`) and accessing the CLI from that port
- [X] Yes, I am accessing the CLI from port `:4280`

**Describe the bug**
When starting an app with `swa start` I get `WARNING: Failed to load staticwebapp.config.json schema. Continuing without validation!`.
Even if it's just a warning and is referring to the schema file (not the config itself) `staticwebapp.config.json` doesn't work as expected. e.g. none of the routes declared is working.

**To Reproduce**

1. Go to any swa app
2. Type in command 'swa start'
3. See warning in console
4. Check in browser any routes declared -> won't work.

**Expected behavior**
 `staticwebapp.config.json` to work. Routes and any other settings declared to work.

**Desktop (please complete the following information):**

- OS: Windows 
- Version: 11

**Additional context**
After investigating this issue i have discovered that the URL used to load the config schema is hardcoded: https://github.com/Azure/static-web-apps-cli/blob/affefb24872286a542fe7bbd54340648ec3294d0/src/core/utils/user-config.ts#L168C13-L168C13 .
At this moment, the website https://www.schemastore.org/ which is hosting the schema file, is down (error 503).
Allowing loading the schema file only from a single website has created a single point of failure which we are experiencing right now.

`staticwebapp.config.json` has a property `$schema`, but as you can see the value is not used for actually loading the schema.
My suggestion is to allow the user to provide a value for `$schema` (in order to no longer depend on https://www.schemastore.org/ ).

Thank you.

Edit: there's currently an issue on Schemastore repository: https://github.com/SchemaStore/schemastore/issues/3303

LiamOSullivan:
I'm also seeing this warning, but routes are loading okay. I am using a customised config file swa-cli-config.json

My redirect from /login causes an error and the swa auth emulator does not seem to be protecting/allowing routes correctly. Trying to determine if related to above issue or not...

abaudhuin:
I have this issue also. It fails silently and won't load my settings. 
I use the config to set CORS, so my deployment is not accessible at all from my app. 
I cannot deploy anymore until I have a solution for this. I'm completely blocked.

Please, do you have a solution ?

froleo:
[...]

PR #803, opened 2024-03-04
add config schema option

imcat0131:
How about adding an option to specify the JsonSchema distribution part, which often causes problems in issues like the one below.

Issue: https://github.com/Azure/static-web-apps-cli/issues/754

I imagine the following command as how to use it.

```bash
swa start --schema "https://[alterbaseurl]/staticwebapp.config.json"
```


imcat0131:
I am very grateful for this resource.
I would like to return the favor to your team.
If there is a problem, I will fix it immediately.

imcat0131:
@microsoft-github-policy-service agree

imcat0131:
@microsoft-github-policy-service agree

Issue #802, opened 2024-03-04
Dependency on json.schemastore.org breaks CLI


itpropro:
The SWA CLI currently fails and stops after a few seconds, because either the latency of https://json.schemastore.org/staticwebapp.config.json is too high or the service is not available.
There should be no dependency on an online resource to be available for local development. From the logs it seems like it should continue to work without validation, but it currently doesn't.

Logs:

```
[swa] network timeout at: https://json.schemastore.org/staticwebapp.config.json
[swa] ✖ Failed to load staticwebapp.config.json schema. Continuing without validation!
[swa] node "C:\Users\USER\AppData\Local\pnpm\global\5\.pnpm\@azure+static-web-apps-cli@1.1.6\node_modules\@azure\static-web-apps-cli\dist\msha\server.js" exited with code 0
--> Sending SIGTERM to other processes..
[dataApi] cd "C:\repos\APP\data-api" && "C:\Users\USER\.swa\dataApiBuilder\0.9.7\Microsoft.DataApiBuilder.exe" start -c staticwebapp.database.config.json --no-https-redirect exited with code 1
--> Sending SIGTERM to other processes..
[run] cd "C:\repos\APP\app" && pnpm dev exited with code 1
✖ SWA emulator stopped because the --run command exited with code 1.
```

My suggestion would be to increase the timeout for the request and to fix the bug that stops the CLI from working without validation.

cristianbretti:
Heavy plus one! We use this CLI deploy our production sites, so this is really blocking us.

It is this line that makes this fail. It should not send `true` as the second argument as `true` is used to signal an exit. 

https://github.com/Azure/static-web-apps-cli/blob/f9f9b7e92f5fb03cf9b75626377f951a5f85cd5d/src/core/utils/user-config.ts#L108

sandstreampop:
Hi all, we're having troubles with this in a production project as well.

As a workaround in our project, we've put the schema right into the source code to remove the dependency to this online resource. 

In node_modules/@azure/static-web-apps-cli/dist/core/utils/user-config.js, line 151, we've changed the `loadSWAConfigSchema` function as follows:

```js
async function loadSWAConfigSchema() {
    const schemaUrl = "https://json.schemastore.org/staticwebapp.config.json";
    try {
        // ORIGINAL CODE WHICH CURRENTLY FAILS DUE TO ONLINE DEPENDENCY:
[...]

PR #808, opened 2024-03-04
fix: load staticwebapp.config.json schema locally when timeout

cjk7989:
Added a logic to load static-web-apps-cli/schema/staticwebapp.config.json as a backup to solve issue #754 and #802

### DIFF
--- a/schema/staticwebapp.config.json
+++ b/schema/staticwebapp.config.json
@@ -1,0 +1,613 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "additionalProperties": false,
+  "default": {
+    "navigationFallback": {
+      "rewrite": "/index.html"
+    }
+  },
+  "definitions": {
+    "route": {
+      "type": "object",
+      "required": ["route"],
+      "properties": {
+        "route": {
+          "type": "string",
+          "description": "Request route pattern to match. May contain valid wildcards. See documentation: https://aka.ms/swa/config-schema"
+        },
+        "methods": {
+          "type": "array",
+          "description": "Request method(s) to match",
+          "items": {
+            "anyOf": [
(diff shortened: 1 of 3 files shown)


schema/staticwebapp.config.json
```diff
@@ -0,0 +1,613 @@
...
+      "description": "Platform configuration",
+      "properties": {
+        "apiRuntime": {
+          "type": "string",
+          "enum": [
+            "dotnet:3.1",
+            "dotnet:6.0",
+            "dotnet-isolated:6.0",
+            "dotnet-isolated:7.0",
+            "node:12",
```
pocki, line 592: Latest platforms seems to be missing compared to Azure Functions: 
`az functionapp list-runtimes --os linux --query "[].{stack:join(':', [runtime, version]), LinuxFxVersion:linux_fx_version, SupportedFunctionsVersions:to_string(supported_functions_versions[])}" --output table`

cjk7989: Will investigate the details, thanks!

anfibiacreativa:
Hi @cjk7989 although I agree that this PR at least decouples concerns: 1 thing is not to find the schema, another thing is for the configuration not to conform to the schema. 

Can we please revert the previous changes https://github.com/Azure/static-web-apps-cli/commit/f07e351e5c693d482061829119bcd1f37b92abda on the grounds of: do we really want to block customers from running their applications when there is a misconfiguration? What is the rationale to go from warning to error?

schema/staticwebapp.config.json
cjk7989: Have updated supported platform "dotnet-isolated:8.0" in PR https://github.com/Azure/static-web-apps-cli/pull/809.

PR #817, opened 2024-03-19
fix: Allow SKU selection when creating a new site on "swa deploy"

mkarmark:
Currently, when you create a new site via "swa deploy", the CLI always creates a "Free" sku site, but this PR introduces a sku picker that defaults to Free, but allows SKU selection during site creation. This is especially helpful to those who have hit the 5 free site limit and want to create a new site.


cjk7989 (approved):
approved

PR #826, opened 2024-04-25
feat(auth): Add custom auth support for Google and GitHub providers

andimarc:
### Setup

The user must first create the necessary GitHub OAuth app and/or Google OAuth Client ID with the redirect URIs set to http://localhost:4280/.auth/login/github and http://localhost:4280/.auth/login/google respectively.

They must then add the "auth" configuration in their staticwebapp.config.json file and set the appropriate environment variables containing client ID and client secret values before running starting the SWA CLI.

### GitHub Login

User clicks on "Log in with GitHub" link (http://localhost:4280/.auth/login/github), which redirects to https://github.com/login/oauth/authorize.

AppPage

User logs into GitHub if not already logged in.

GitHubLogin

User authorizes GitHub OAuth app if not already authorized for their account.

GitHubAuthorize

User is redirected back to the static web app.

GitHubLoggedIn

### Google Login

User clicks on "Log in with Google" link (http://localhost:4280/.auth/login/google), which redirects to https://accounts.google.com/o/oauth2/v2/auth.

AppPage

User logs into Google if not already logged in.

GoogleLogin

User authorizes Google OAuth app if not already authorized for their account.

GoogleAuthorize

User is redirected back to the static web app.

GoogleLoggedIn

### DIFF
--- a/src/msha/auth/routes/auth-login-provider-callback.ts
+++ b/src/msha/auth/routes/auth-login-provider-callback.ts
@@ -4,7 +4,7 @@ import * as https from "https";
 import * as querystring from "querystring";
 import { SWA_CLI_API_URI, SWA_CLI_APP_PROTOCOL } from "../../../core/constants";
 import { DEFAULT_CONFIG } from "../../../config";
-import { hashStateGuid } from "../../../core/utils/auth";
+import { hashStateGuid, isNonceExpired } from "../../../core/utils/auth";
 
 const getGithubAuthToken = function (codeValue: string, clientId: string, clientSecret: string) {
   const data = querystring.stringify({
@@ -401,6 +401,16 @@ const httpTrigger = async function (context: Context, request: http.IncomingMess
     return;
   }
 
+  if (isNonceExpired(nonce)) {
+    context.res = response({
+      context,
+      status: 401,
+      headers: { ["Content-Type"]: "text/plain" },
+      body: "Login timed out. Please try again.",
+    });
+    return;
+  }
+
   const { clientIdSettingName, clientSecretSettingName } = customAuth?.identityProviders?.[providerName]?.registration || {};
 
   if (!clientIdSettingName) {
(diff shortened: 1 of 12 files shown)


src/msha/auth/routes/auth-login-provider-callback.ts
```diff
@@ -0,0 +1,480 @@
...
+    });
+    return;
+  }
+
+  const url = new URL(request.url!, `${SWA_CLI_APP_PROTOCOL}://${request?.headers?.host}`);
+
+  const codeValue = url.searchParams.get("code");
+  const stateValue = url.searchParams.get("state");
+
+  const nonce = decodeNonceCookie(cookie);
```
mkarmark, line 389: the typical purpose of a nonce is to validate that it is what we get back at the end of the request as what we started with at the beginning of the loop, right? Does this achieve that? I may be confused but I'm not following what exact validation we are doing here

andimarc: This is verifying that the /callback call corresponds to the provider auth call that we initiated.

Before we redirect to https://github.com/login/oauth/authorize or https://accounts.google.com/o/oauth2/v2/auth, we generate a GUID, which we set in the "Nonce" cookie. We also hash this nonce and pass it as the "state" query string parameter in the provider auth URI that we're redirecting to.

Once the user has logged into the provider, the provider redirects them to our /callback API, passing "code" and "state" as query string parameters to the API. For "state" the provider is passing the same value that we originally passed when redirecting to the provider auth API.

So, if we receive a request to our /callback API, the value of the "state" query string parameter passed to the API should be the hash of the value in the "Nonce" cookie. We verify this by hashing the "Nonce" cookie value and comparing it to the "state" value. If the values don't match (or there is no "Nonce" cookie present), then we know this /callback call doesn't correspond to a provider auth redirect that we initiated and we fail the request with status 401.

mkarmark (approved):

Issue #829, opened 2024-05-15
SWA broken as of node version `20.13.0`


cr3wdayt5p:
**Describe the bug**
The SWA CLI does not work with node version `20.13.0` or later (also tested with `20.13.1` and `22.1.0`). 

There are no issues with node `20.12.2`.

Running `swa start` does not report any errors. However after succesful startup of `swa start` the URL `https://127.0.0.1:8082` (or `https://localhost:8082`) simply refuse connections (Chrome report `ERR_CONNECTION_REFUSED`).

**To Reproduce**

1. With node `20.12.2` setup some SWA project. Verify that `swa start` work as expected.
2. Upgrade to node `20.13.0`. SWA is no longer working (but `swa start` report no errors).

**Expected behavior**
Nothing should break upgrading from node `20.12.2` to `20.13.0` (or later).

**Desktop (please complete the following information):**

- OS: macOS 14.4 (arm64)
- SWA CLI version: `1.1.7`

**Additional context**

`swa-cli.config.json`:
```json
{
  "configurations": {
    "app": {
      "appDevserverUrl": "https://127.0.0.1:8081",
      "host": "127.0.0.1",
      "port": 8082,
      "ssl": true,
      "sslCert": "localhost-cert/localhost.crt",
      "sslKey": "localhost-cert/localhost.key"
    }
  }
}
```

bzbetty:
can reproduce.

seems that it's binding to the external ip of the local machine.

codingoutloud:
Also reproduced. 

**swa --version =** 
1.1.8

**sw_vers =** 
ProductName:            macOS
ProductVersion:         14.5
BuildVersion:           23F79

yamachu:
I investigated and it seems that there has been a change in the behavior of node's http module.
ref: https://github.com/nodejs/node/issues/53204

There are two ways to deal with this.
The first is to access it with a localIpAdress address. (e.g. 192.168.xxx.xxx)
The second is to remove the following line (I would like to ask for instructions from the product owner :eyes:)
https://github.com/Azure/static-web-apps-cli/blob/352be8f4ce2d01e1dac17797a80f9f414d612bc0/src/msha/server.ts#L174

If you do either of these, it looks like you can get it to work with

johnnyreilly:
@yamachu can you expand on what you mean by:

> The first is to access it with a localIpAdress address. (e.g. 192.168.xxx.xxx)

please?

yamachu:
[...]

PR #841, opened 2024-05-30
Remove the second server.listen so that it also works with Node 20.13

yamachu:
ref: https://github.com/Azure/static-web-apps-cli/issues/829

Since Node 20.13, the behavior of `Server.listen` has changed.
The last listen will activate.
see: https://github.com/nodejs/node/issues/53204

The SWA CLI used to listen for localhost and listen for localIpAdress (access from the same network), but in Node 18 and 20, the latter listen for localIpAdress does not seem to work anymore.
Only localhost listen is performed under 20.13.

Currently only the latter process is used in Node 20.13 environments, so only localIpAdress access is available.

Therefore, this PR will be adapted to the behavior under Node 20.13 so that it can be accessed by localhost and so on.

I think it is better to consider how to make it accessible via localIpAdress from the same network in a new PR.

Testing with this change can be verified with `npm run e2e:static`. (This is the e2e test that is currently disabled)

### DIFF
--- a/src/msha/server.ts
+++ b/src/msha/server.ts
@@ -3,7 +3,6 @@ import fs from "fs";
 import http from "http";
 import httpProxy from "http-proxy";
 import https from "https";
-import internalIp from "internal-ip";
 import net from "net";
 import open from "open";
 import { DEFAULT_CONFIG } from "../config";
@@ -146,7 +145,6 @@ function onServerStart(server: https.Server | http.Server, socketConnection: net
 // start SWA proxy server
 (async () => {
   let socketConnection: net.Socket | undefined;
-  const localIpAdress = await internalIp.v4();
 
   // load user custom rules if running in local mode (non-dev server)
   let userConfig: SWAConfigFile | undefined;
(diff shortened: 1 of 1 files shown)


Timothyw0:
@yamachu Thank you for looking into this issue and identifying the fix! I just tested in my local env with Node 18 and 20 and it's working great!

IvanJobs:
Thanks for identifying and fixing this issue.

I'm wondering if there's some better solution for this. How about we make it listening on two endpoints for real by using two server instances?
Check https://www.quora.com/How-do-I-host-a-Node-js-app-that-listens-on-multiple-ports.

johnnyreilly:
Why would that be a better solution? If this fixes the issue then I'm very happy with that as a result!

IvanJobs:
> Why would that be a better solution? If this fixes the issue then I'm very happy with that as a result!
Just need to make sure it will not introduce another issue based on the original behavior which is listening on two endpoints.
To be robust, this can split into two steps:
1. Keep the original behavior and fix the issue.
2. If listening on two endpoints is useless, we can remove one of them in another PR.
That's my opinion for reference.

IvanJobs (approved):
Confirmed with the original code owner and the team, it's good to just listen on one endpoint, so approve it.

Timothyw0 (approved):

PR #880, opened 2024-08-13
[Feature] Add Facebook Custom Auth Support

Timothyw0:
### Description
New feature: Adding Facebook custom auth support using Facebook's manual OIDC token flow. A few more refactors had to be done to the previous code since Facebook's flow is a little bit different, mainly: 
* Facebook's auth provider schema is different in staticwebapp.config.json
* There is no introspection endpoint after getting the token so we need to manually decode the returned JWT

### Testing
staticwebapp.config.json:
image

Facebook /.auth/me:
image

Github /.auth/me:
image

AAD /.auth/me:
image

Google /.auth/me:
image

### DIFF
--- a/src/msha/auth/index.ts
+++ b/src/msha/auth/index.ts
@@ -25,7 +25,7 @@ function getAuthPaths(isCustomAuth: boolean): Path[] {
     paths.push({
       method: "GET",
       // For providers with custom auth support not implemented, revert to old behavior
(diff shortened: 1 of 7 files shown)


src/msha/auth/index.ts
cjk7989, line 1: For the comments in Line 15:
// only match for providers with custom auth support implemented (github, google, aad)

Should we update the supported provider type?

cjk7989 (approved):
It looks that the PR could work well in both Windows and Linux. Approved.

PR #889, opened 2024-09-26
[Feature] Add Twitter/X Custom Auth Support

Timothyw0:
### Description
New feature: Adding Twitter/X custom auth support using Twitter's manual OAuthv2 token flow.
Some quirks of Twitter's flow:
* It is not an OIDC flow so the returned information does not encompass as much as other providers
* Different properties of client id and secret in staticwebappconfig.json schema

### Additional Fix
Chore: Updated actions/upload-artifact version since v2 is now deprecated
image

### Testing
Twitter staticwebappconfig.json:
image

Twitter /.auth/me:
image

Regression Testing below

AAD/Entra ID /.auth/me:
image

Google /.auth/me:
image

Facebook /.auth/me:
image

Github /.auth/me:
image

### DIFF
--- a/src/core/constants.ts
+++ b/src/core/constants.ts
@@ -50,7 +50,7 @@ export const SWA_AUTH_COOKIE = `StaticWebAppsAuthCookie`;
 export const ALLOWED_HTTP_METHODS_FOR_STATIC_CONTENT = ["GET", "HEAD", "OPTIONS"];
 
 // Custom Auth constants
-export const SUPPORTED_CUSTOM_AUTH_PROVIDERS = ["google", "github", "aad", "facebook", "dummy"];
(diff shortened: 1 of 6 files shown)


src/core/constants.ts
```diff
@@ -73,6 +73,10 @@ export const CUSTOM_AUTH_TOKEN_ENDPOINT_MAPPING: AuthIdentityTokenEndpoints = {
     host: "graph.facebook.com",
     path: "/v11.0/oauth/access_token",
   },
+  twitter: {
+    host: "api.twitter.com",
```
cjk7989, line 77: It looks good to me. But where do you get the authentication endpoints for the providers?

cjk7989 (approved):

PR #892, opened 2024-10-17
(fix #890) Fix Undefined Error with No Subs in Azure Profile

### LINKED ISSUE #890: [Bug] CLI assumes the AZURE_LOGIN_CONFIG file has subscriptions in it
I'm using the latest version 2.0.1

**Describe the bug**
CLI is failing to deploy when reading the contents of my AZURE_LOGIN_CONFIG file. Logs below:

```
✖ Cannot read properties of undefined (reading 'find')
✔ Successfully logged into Azure!
Found an existing Azure config file, getting Tenant and Subscription Id from /home/vsts/.azure/azureProfile.json
```

This is the content of my `azureProfile.json`: 

```json
{"installationId": "redacteduuid"}
```

I could track down the error to be on this statement:

https://github.com/Azure/static-web-apps-cli/blob/5f440e3068be7d53e052fae17e58dcc5c8b108e5/src/cli/commands/login/login.ts#L165

And the line before coerces the configuration to an interface: 

https://github.com/Azure/static-web-apps-cli/blob/5f440e3068be7d53e052fae17e58dcc5c8b108e5/src/cli/commands/login/login.ts#L164

Which resolves to this interface:

https://github.com/Azure/static-web-apps-cli/blob/5f440e3068be7d53e052fae17e58dcc5c8b108e5/src/swa.d.ts#L443-L446

Which is not true. My `azureProfile.json` doesn't have the subscriptions key. Therefore the failure.

**To Reproduce**

1. Make a pipeline in Azure Dev Ops
2. Use a service principal logging in with client ID and client secret in an AzureCLI@2 task
3. Inside of an AzureCLI@2 task, run the SWA CLI commands via an inline script
4. See error. The command will not fail despite the failure which makes debugging very hard once CI will never fail.

**Expected behavior**

Expected to have the command working fine.

Timothyw0:
Make the `subscriptions` property optional in AzureProfile since it may not exist as described in Issue #890. 

Tested by removing subscriptions in my azureProfile.json and code will no longer throw an undefined error. Then the login will fallback to another means to authenticate.

### DIFF
--- a/src/cli/commands/login/login.ts
+++ b/src/cli/commands/login/login.ts
@@ -56,7 +56,7 @@ async function setupProjectCredentials(options: SWACLIConfig, credentialChain: T
     const tenants = await listTenants(credentialChain);
     if (tenants.length === 0) {
       throw new Error(
-        `No Azure tenants found in your account.\n  Please read https://docs.microsoft.com/azure/cost-management-billing/manage/troubleshoot-sign-in-issue`
+        `No Azure tenants found in your account.\n  Please read https://docs.microsoft.com/azure/cost-management-billing/manage/troubleshoot-sign-in-issue`,
       );
     } else if (tenants.length === 1) {
       logger.silly(`Found 1 tenant: ${tenants[0].tenantId}`);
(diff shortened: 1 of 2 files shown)


cjk7989 (approved):
The PR seems well-implemented for the described issue

Issue #970, opened 2025-08-06
Version incompatibility for Azure Functions Core Tools blocks execution of `swa start`


stephlocke:
**Are you accessing the CLI from the default port `:4280` ?**

- [ ] No, I am using a different port number (`--port`) and accessing the CLI from that port
- [x] Yes, I am accessing the CLI from port `:4280`

**Describe the bug**
Working in ubuntu on the latest version of node and the swa cli, adding managed apis causes the cli to no longer work in terms of starting the app because of a version incompatability being registered between node and the azure functions core tools.

> ✖ Found Azure Functions Core Tools v4 which is incompatible with your current Node.js v24.5.0.
> ✖ See https://aka.ms/functions-node-versions for more information.

**To Reproduce**

1. Install node
```
> node --version
v24.5.0
```
2. Install swa 
```
> swa --version
2.0.6
```
3. In an swa project, create a dotnet8 isolated function runtime (could work with others but this is what I used) 
4. Run `swa start` triggering core tools grab and then error due to version incompatability
```
>swa start
✖ Found Azure Functions Core Tools v4 which is incompatible with your current Node.js v24.5.0.
✖ See https://aka.ms/functions-node-versions for more information.
```

**Expected behavior**
`swa start` should work OR I should be able to disable parts of the project from being booted up to get around these sorts of errors.

**Desktop (please complete the following information):**

- Ubuntu 24.04

unowiz:
@stephlocke I've the same problem and I switched to use node 22. Check this https://github.com/Azure/static-web-apps-cli/issues/791#issuecomment-2103232652

The cli seems to restricts itself to lts version
`
export function isCoreToolsVersionCompatible(coreToolsVersion, nodeVersion) {
    // Runtime support reference: https://docs.microsoft.com/azure/azure-functions/functions-versions?pivots=programming-language-javascript#languages
    switch (coreToolsVersion) {
        case 4:
            return nodeVersion >= 18 && nodeVersion <= 22;
        case 3:
            return nodeVersion >= 14 && nodeVersion <= 20;
        case 2:
            return nodeVersion >= 10 && nodeVersion <= 14;
        default:
            return false;
    }
}
`

Shuvra17:
Thanks for working on this issue. We’re currently facing : Azure Functions Core Tools v4 is not compatible with newer Node versions (v22/v24), which is impacting our local SWA development and build flows.
[...]

PR #971, opened 2025-08-07
fix: Increment supported version #970

stephlocke:
In May, node released version 24.* - whilst this is not documented as supported yet, testing with version 4.1.10 of the core tools by npm and following the same sort of process swa cli takes, the `func` runs without error.

The range checkers currently prevent managed function dev for people with up-to-date environments.

### DIFF
--- a/src/core/func-core-tools.ts
+++ b/src/core/func-core-tools.ts
@@ -42,7 +42,7 @@ export function isCoreToolsVersionCompatible(coreToolsVersion: number, nodeVersi
   // Runtime support reference: https://docs.microsoft.com/azure/azure-functions/functions-versions?pivots=programming-language-javascript#languages
   switch (coreToolsVersion) {
     case 4:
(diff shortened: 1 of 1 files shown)


LongOddCode:
LGTM for 2.0.9 inclusion — the change is a clean 2-line range extension in `isCoreToolsVersionCompatible` / `detectTargetCoreToolsVersion` (Node 22 → 24). This unblocks developers on current LTS/non-LTS Node when using managed-function workflows, which currently fail range validation even though `func-core-tools` 4.1.10+ runs fine (per @stephlocke's testing).

Suggestions for follow-up:
- Consider bumping the upper bound to **26** to cover Node 26 (current release line at time of this comment) in the same PR, since the func-core-tools behavior is the same and it avoids another range-extension PR next cycle.
- The inline comment still links to the old `docs.microsoft.com` URL — worth updating to `learn.microsoft.com`, but not blocking.

(Drive-by review from a fellow contributor; maintainers own the final call.)

Issue #952, opened 2025-04-13
API with Python 3.11 runtime fails "swa start"


chrisred:
Before filing this issue, please ensure you're using the latest CLI by running `swa --version` and comparing to the latest version on [npm](https://www.npmjs.com/package/@azure/static-web-apps-cli).

**Are you accessing the CLI from the default port `:4280` ?**

- [ ] No, I am using a different port number (`--port`) and accessing the CLI from that port
- [X] Yes, I am accessing the CLI from port `:4280`

> Make sure you are accessing the URL printed in the console when running `swa start`!

**ℹ️ NOTE: Make sure to enable debug logs when running any `swa` commands using `--verbose=silly`**

**Describe the bug**

**To Reproduce**

1. Create a SWA project using a Python 3.11 API runtime
2. Build the project with `swa build my-project`
3. Start with `swa start ./dist --api-location api`
4. `swa start` fails with "Failed to validate staticwebapp.config.json schema."

**Expected behavior**
`swa start` should start the emulator successfully.

**Desktop (please complete the following information):**

- OS: Fedora
- Version 40

**Additional context**

Looks like the file at `schema/staticwebapp.config.json` doesn't match the list of supported runtimes at https://learn.microsoft.com/en-gb/azure/static-web-apps/configuration#select-the-api-language-runtime-version (`python:3.11` is missing).

```
✖ Failed to validate staticwebapp.config.json schema. Errors: [
  {
    "type": "error",
    "name": "EnumError",
    "code": "enum-error",
    "message": "Expected given value `python:3.11` in `#/platform/apiRuntime` to be one of `[\"dotnet:3.1\",\"dotnet:6.0\",\"dotnet-isolated:6.0\",\"dotnet-isolated:7.0\",\"dotnet-isolated:8.0\",\"dotnet-isolated:9.0\",\"node:12\",\"node:14\",\"node:16\",\"node:18\",\"node:20\",\"python:3.8\",\"python:3.9\",\"python:3.10\"]`",
    "data": {
      "pointer": "#/platform/apiRuntime",
      "schema": {
        "type": "string",
        "enum": [
          "dotnet:3.1",
          "dotnet:6.0",
          "dotnet-isolated:6.0",
          "dotnet-isolated:7.0",
          "dotnet-isolated:8.0",
          "dotnet-isolated:9.0",
          "node:12",
          "node:14",
          "node:16",
          "node:18",
          "node:20",
          "python:3.8",
          "python:3.9",
[...]

PR #972, opened 2025-08-08
Increment python version compatability #952 

stephlocke:
Per https://devguide.python.org/versions/ currently released python version is 3.13 with a new one out in October


stephlocke:
@microsoft-github-policy-service agree

PR #989, opened 2026-04-09
fix: propagate non-zero exit codes from deploy client

### LINKED ISSUE #665: Getting zero exit code on deploy failure
**Describe the bug**
There is an upload failure, but CLI returns 0 exit code.

**To Reproduce**

1. Given issue with SWA Azure - like this one => https://github.com/Azure/static-web-apps/issues/1097
2. When running the command 'swa deploy --env production'
3. Then I can see error in log 
>✖ Deployment Failed :(
>✖ Deployment Failure Reason: Failed to deploy the Azure Functions.
 and exit code is "0"

**Expected behavior**
Have non-zero exit code, to get failures in pipelines

**Screenshots**
2023-03-08 non-zero

- OS: Debian GNU/Linux 11 (bullseye)
- Version 1.0,6

LongOddCode:
## Summary

Fixes #665

The deploy() function spawned the deploy client binary (StaticSitesClient) but never awaited the child process. The close event handler ran in a fire-and-forget callback, so a non-zero exit code was silently ignored and the function always resolved — making CI/CD pipelines believe deployment succeeded even when it failed.

### src/cli/commands/deploy/deploy.ts
- Wrap the spawn() call and its event handlers in new Promise and await it
- Reject the promise on non-zero exit codes with a descriptive error message
- Reject on signal kills (e.g., SIGTERM)
- Reject on spawn errors (e.g., ENOENT when binary not found)
- Add a settled guard to prevent double-resolution edge cases
- Re-throw caught errors in the catch block (previously swallowed silently)
- Moved cleanUp() out of the close handler into the existing finally block

### src/cli/commands/deploy/deploy-exit-code.spec.ts (NEW)
- 6 focused unit tests for exit code propagation:
  - Rejects on non-zero exit code
  - Resolves on exit code 0
  - Rejects when killed by signal
  - Rejects on spawn error events
  - Verifies cleanUp() is called after failure
  - Verifies logGitHubIssueMessageAndExit() is called on failure

### DIFF
--- a/src/cli/commands/deploy/deploy.spec.ts
+++ b/src/cli/commands/deploy/deploy.spec.ts
@@ -1,5 +1,7 @@
 import "../../../../tests/_mocks/fs.js";
-import child_process from "node:child_process";
+import child_process, { spawn } from "node:child_process";
+import { EventEmitter } from "node:events";
+import path from "node:path";
 import { logger } from "../../../core/utils/logger.js";
 import { vol } from "memfs";
 import * as accountModule from "../../../core/account.js";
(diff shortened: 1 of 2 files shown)


cjk7989 (commented):
Hi @LongOddCode, the logic looks correct and I would approve it. Could you attach the test result?

cjk7989 (approved):

PR #999, opened 2026-06-05
fix: validate --env is not passed with an empty string

monteiro-renato:
Looks like the CLI accepts --env="", and the deployment seems to "work"; There were no error messages in the console at least.
Looking at the portal it doesn't seem like anything gets deployed however.

I am not exactly sure how I could dig deeper into what happens when the deployment code is actually running, but I guess adding this extra validation can't hurt.

### DIFF
--- a/src/cli/commands/deploy/deploy.spec.ts
+++ b/src/cli/commands/deploy/deploy.spec.ts
@@ -313,7 +313,7 @@ describe("deploy", () => {
         outputLocation: "/test-output",
         env: "",
       });
-      expect(logger.error).toHaveBeenNthCalledWith(1, "Invalid --env: cannot be empty. Use 'preview' or omit the flag.");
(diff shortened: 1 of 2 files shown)


cjk7989:
Hi @monteiro-renato , thanks for the fix! The bug is real and worth addressing. A few concerns:

### Validation fires too late

By the time this check runs (~line 258), line 190 has already printed:
`Deploying to environment: `

The validation should be moved to the top of the function alongside other input checks, before any logging or processing occurs.

### Error message is slightly misleading

Users can specify any custom environment name, not just preview. Suggest:

"Use a named environment (e.g. 'preview') or omit the flag."

monteiro-renato:
@cjk7989, thanks for the review!
I've moved it upwards, let me know if there is anything else you would like me to change.

cjk7989 (approved):

cjk7989:
Looks good to me, approved. Thank you again :)

PR #1001, opened 2026-07-12
fix supported versions

svrooij:
It is a mistery to my why the cli is only supporting very old versions. Here is a version bump for node an `dotnet-isolated` which for some reason are referenced as `dotnetisolated` only in this repository.

And you might want to change the config specification to also reflect this..... [schemastore.org/staticwebapp.config.json](https://www.schemastore.org/staticwebapp.config.json)

image

```json
{
    "platform": {
      "type": "object",
      "description": "Platform configuration",
      "properties": {
        "apiRuntime": {
          "type": "string",
          "enum": [
            "dotnet:3.1",
            "dotnet:6.0",
            "dotnet-isolated:6.0",
            "dotnet-isolated:7.0",
            "dotnet-isolated:8.0",
            "dotnet-isolated:9.0",
            "node:12",
            "node:14",
            "node:16",
            "node:18",
            "node:20",
            "python:3.8",
            "python:3.9",
            "python:3.10"
          ],
          "description": "Language runtime for the managed functions API"
        }
      },
      "additionalProperties": false
    },
}
```

### DIFF
--- a/schema/staticwebapp.config.json
+++ b/schema/staticwebapp.config.json
@@ -588,6 +588,7 @@
             "dotnet:8.0",
             "dotnet-isolated:8.0",
             "dotnet-isolated:9.0",
+            "dotnet-isolated:10.0",
             "node:18",
             "node:20",
             "node:22",
--- a/src/cli/commands/deploy/deploy.ts
+++ b/src/cli/commands/deploy/deploy.ts
@@ -37,6 +37,11 @@ export async function deploy(options: SWACLIConfig) {
     apiVersion,
   } = options;
 
+  // normalize the hyphenated variant to the canonical value used throughout the codebase
+  if (apiLanguage === "dotnet-isolated") {
+    apiLanguage = "dotnetisolated";
+  }
+
   if (dryRun) {
     logger.warn("***********************************************************************");
     logger.warn("* WARNING: Running in dry run mode. This project will not be deployed *");
--- a/src/core/constants.ts
+++ b/src/core/constants.ts
@@ -305,7 +305,7 @@ export const DEFAULT_VERSION = {
 };
 
 export const SUPPORTED_VERSIONS = {
-  Node: ["18", "20", "22", "24", "26"],
+  Node: ["18", "20", "22"],
   Dotnet: ["8.0"],
   DotnetIsolated: ["8.0", "9.0", "10.0"],
   Python: ["3.9", "3.10", "3.11"],
--- a/src/core/constants.ts
+++ b/src/core/constants.ts
@@ -305,9 +305,9 @@ export const DEFAULT_VERSION = {
 };
 
 export const SUPPORTED_VERSIONS = {
-  Node: ["18", "20", "22"],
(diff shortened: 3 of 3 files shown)


simonechGSC:
I had just created a ticket before seeing this PR :) 
https://github.com/Azure/static-web-apps-cli/issues/1002

svrooij:
@simonechGSC just deploy your app with `dotnetisolated` and `8.0` as parameters. I successfully deployed my `dotnet-isolated v10` app using those parameters. Then you can continue with your day while they might accept this pr

simonechGSC:
> @simonechGSC just deploy your app with `dotnetisolated` and `8.0` as parameters. I successfully deployed my `dotnet-isolated v10` app using those parameters. Then you can continue with your day while they might accept this pr

That's indeed what I did... I wonder what is the purpose of that parameter if is not taken into account

svrooij:
Great question for the team that developed this cli.
I'm wondering the same, and I'm glad I found a way that works.

cjk7989:
Hi @svrooij, thank you very much for the contribution. StaticWebApps platform haven't supported node 24, 26 currently, so just making changes on SWA-CLI will not help. We will support them in the future. See https://learn.microsoft.com/en-us/azure/static-web-apps/languages-runtimes

simonechGSC:
> Hi @svrooij, thank you very much for the contribution. StaticWebApps platform haven't supported node 24, 26 currently, so just making changes on SWA-CLI will not help. We will support them in the future. See https://learn.microsoft.com/en-us/azure/static-web-apps/languages-runtimes

According to that page, net10 is not supported either, but I deploy it, it works fine.
.NET 10 is the currently supported LTS, seems strange SWA are not supporting it yet... is this SWA a "dying" service or is it there to stay?

Also that page was last updated in february 2025, not really up to date
image

svrooij:
@cjk7989 I removed the node versions (that should be supported), and made it accept both `dotnet-isolated` (as any other cli from microsoft) and `dotnetisolated` to keep others from mistakes.

svrooij:
@cjk7989 care to have a look at this PR? It is an easy fix that will unblock others. While not changing anything mayor.

Not officially supporting dotnet 10 really makes users of Swa nervous, and they start asking questions if SWA is still a product Microsoft wants to support.

svrooij:
@cjk7989 @jessieyyt @LongOddCode anyone care to check out this pr? How does this project work?

simonechGSC:
If SWA does not support NET 10, there will soon be no supported version of .NET that can be used for SWA, as 8 and 9 will be out of support in a few months

image

svrooij:
> If SWA does not support NET 10, there will soon be no supported version of .NET that can be used for SWA, as 8 and 9 will be out of support in a few months
> 
> image
> 

☝️ This!!!

Come on Microsoft, please let us have a supported dotnet version in static web apps

svrooij:
Jikun (@cjk7989) jessieyyt Long Hao (@LongOddCode) anyone care to check out this pr?

Ding dong?

PR #1008, opened 2026-08-24
fix(core): chmod the Core Tools binaries that are actually present

### LINKED ISSUE #1007: [Bug] `downloadCoreTools` fails with `ENOENT: chmod 'gozip'` because current Core Tools releases no longer ship `gozip`
Before filing this issue, please ensure you're using the latest CLI by running `swa --version` and comparing to the latest version on [npm](https://www.npmjs.com/package/@azure/static-web-apps-cli).

I am using `@azure/static-web-apps-cli` **2.0.10**, which is the latest version on npm at the time of writing.

**Are you accessing the CLI from the default port `:4280` ?**

- [ ] No, I am using a different port number (`--port`) and accessing the CLI from that port
- [x] Yes, I am accessing the CLI from port `:4280`

(Note: the failure occurs during startup, before the emulator begins listening, so the port is not a factor in this bug.)

**ℹ️ NOTE: Make sure to enable debug logs when running any `swa` commands using `--verbose=silly`**

<details>
<summary>Verbose log (user-specific paths and names redacted as <code>&lt;user&gt;</code>, <code>&lt;project-root&gt;</code>, etc.)</summary>

```
Welcome to Azure Static Web Apps CLI (2.0.10)
Getting config file options from "swa-cli.config.json"...
Config file does not exist at "swa-cli.config.json"
***********************************************************************
* WARNING: This emulator may not match the cloud environment exactly. *
* Always deploy and test your app in Azure.                           *
***********************************************************************
Checking if localhost:4280 is accepting TCP connections...
Port 4280 is available. Use it.
Resolved port number: 4280
appDevserverUrl provided, we will try connect to dev server at .
Api Folder found: /home/<user>/<project-root>/api
Trying to read workflow config with values:
 - appLocation: /home/<user>/<project-root>
 - outputLocation: http://localhost:<dev-server-port>
 - apiLocation: /home/<user>/<project-root>/api
Found a SWA workflow file: /home/<user>/<project-root>/.github/workflows/azure-static-web-apps-<name>.yml
Error reading workflow configuration:
missing property "jobs.build_and_deploy_job" in the SWA workflow file "/home/<user>/<project-root>/.github/workflows/azure-static-web-apps-<name>.yml".
See https://docs.microsoft.com/azure/static-web-apps/build-configuration?tabs=github-actions#build-configuration for more information.
✖ Failed to download Functions Core Tools v4.
✖ Error: ENOENT: no such file or directory, chmod '/home/<user>/.swa/core-tools/v4/gozip'
Error: ENOENT: no such file or directory, chmod '/home/<user>/.swa/core-tools/v4/gozip'
    at Object.chmodSync (node:fs:2050:11)
    at downloadCoreTools (file:///home/<user>/.npm/_npx/<hash>/node_modules/@azure/static-web-apps-cli/src/core/func-core-tools.ts:257:8)
    at processTicksAndRejections (node:internal/process/task_queues:104:5)
    at getCoreToolsBinary (file:///home/<user>/.npm/_npx/<hash>/node_modules/@azure/static-web-apps-cli/src/core/func-core-tools.ts:290:5)
    at start (file:///home/<user>/.npm/_npx/<hash>/node_modules/@azure/static-web-apps-cli/src/cli/commands/start/start.ts:182:26)
    at Command.<anonymous> (file:///home/<user>/.npm/_npx/<hash>/node_modules/@azure/static-web-apps-cli/src/cli/commands/start/register.ts:75:7)
    at Command.parseAsync (/home/<user>/.npm/_npx/<hash>/node_modules/commander/lib/command.js:935:5)
    at run (file:///home/<user>/.npm/_npx/<hash>/node_modules/@azure/static-web-apps-cli/src/cli/index.ts:98:3) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'chmod',
  path: '/home/<user>/.swa/core-tools/v4/gozip'
}
✖ 
Could not find or install Azure Functions Core Tools.
Install Azure Functions Core Tools with:
  npm i -g azure-functions-core-tools@4 --unsafe-perm true
See https://aka.ms/functions-core-tools for more information.
```

Note: the "Error reading workflow configuration" warning above comes from my local workflow file and is unrelated to this bug — the same `ENOENT: chmod 'gozip'` failure reproduces with the minimal steps below, without any workflow file.

</details>

**Describe the bug**

When `swa start` is run with `--api-location`, the CLI calls `getCoreToolsBinary()`, which downloads Azure Functions Core Tools if no usable installation is found. After the package is downloaded and unzipped successfully, `downloadCoreTools()` (`src/core/func-core-tools.ts`) unconditionally runs `chmodSync` on a file named `gozip` on Linux/macOS. Current Core Tools releases no longer include `gozip`, so the call throws `ENOENT` and the entire startup fails — even though the `func` binary itself was downloaded and extracted correctly.

Root cause details are in **Additional context** below.

**To Reproduce**

Steps to reproduce the behavior (minimal reproduction — the failure is independent of any project contents, because it happens inside `downloadCoreTools` before the API is ever started):

1. Use a Linux x64 machine (Ubuntu, Node.js v24) where the `func` binary is not detected globally
2. Remove any cached Core Tools: `rm -rf ~/.swa/core-tools`
3. Create an empty directory to pass as the API location: `mkdir -p app api`
4. Run: `npx @azure/static-web-apps-cli@2.0.10 start ./app --api-location ./api`
5. The CLI attempts to download Functions Core Tools v4 and fails with `ENOENT: no such file or directory, chmod '~/.swa/core-tools/v4/gozip'`

**Expected behavior**

- Startup should succeed as long as `func` itself was downloaded and extracted correctly, even if `gozip` is absent. For the local development scenario covered by `swa start`, `gozip` was a packaging/deployment helper and should not be required.
- At minimum, the code should check whether the file exists (e.g. `fs.existsSync`) before calling `chmodSync`, and skip (or log a warning) if it does not.

**Screenshots**

N/A — see the error log above.

- OS: Ubuntu (Linux x64)
- Node.js: v24.18.0
- `@azure/static-web-apps-cli`: 2.0.10
- `azure-functions-core-tools` (installed via npm, for reference): 4.13.0

**Additional context**

*Root cause (investigated):*

`gozip` is **no longer included** in current `azure-functions-core-tools` releases — neither in the npm distribution (4.13.0) nor in the full `linux-x64` archive (4.12.1) downloaded directly from GitHub Releases. I inspected the extracted archives: there is no file named `gozip`; instead they ship `System.IO.Compression.ZipFile.dll` (the standard .NET compression library). It appears the Core Tools team replaced the external Go binary (`gozip`) used for packaging with the standard .NET library.

However, `downloadCoreTools` still assumes `gozip` exists and chmods it unconditionally on Linux/macOS (current `main`, `src/core/func-core-tools.ts`):

```ts
  // Fix permissions on MacOS/Linux
  if (os.platform() === "linux" || os.platform() === "darwin") {
    fs.chmodSync(path.join(dest, "func"), 0o755);
    fs.chmodSync(path.join(dest, "gozip"), 0o755); // <-- throws ENOENT
  }
```

*Suggested fix:*

Guard the `chmodSync` call on `gozip` with an existence check:

```ts
const gozipPath = path.join(dest, "gozip");
if (fs.existsSync(gozipPath)) {
  fs.chmodSync(gozipPath, 0o755);
}
```

*Related issue:*

#899 is a past case where the download flow broke because assumptions about the distribution feed/artifacts no longer held (Windows, missing `size: "full"` artifact). This report is a different, newer cause: the removal of `gozip` itself from the shipped artifacts.

*Workaround:*

Skip the automatic Core Tools startup entirely: start Azure Functions Core Tools separately with `func start`, then connect the SWA CLI to it:

```
swa start <app-url> --api-devserver-url <local-api-url>
```

This corresponds to the "Manual start" flow in the [official docs](https://learn.microsoft.com/azure/static-web-apps/static-web-apps-cli-api-server#manual-start) and does not go through `downloadCoreTools`, so it is unaffected by this bug.

Om-singhaI:
# fix(core): chmod the Core Tools binaries that are actually present

Fixes #1007

## What breaks

On Linux and macOS, a fresh `swa start` that has to download Azure Functions Core Tools v4 dies immediately after the archive extracts successfully:

```
Error: ENOENT: no such file or directory, chmod '~/.swa/core-tools/v4/gozip'
```

`downloadCoreTools()` in `src/core/func-core-tools.ts` extracts the package, then unconditionally runs chmod on two fixed paths:

```ts
// Fix permissions on MacOS/Linux
if (os.platform() === "linux" || os.platform() === "darwin") {
  fs.chmodSync(path.join(dest, "func"), 0o755);
  fs.chmodSync(path.join(dest, "gozip"), 0o755);
}
```

The second call is the crash. Nothing catches it, so the CLI fails at startup even though the download and extraction both worked.

## Why it happens

`gozip` was not removed from Core Tools. It moved.

I checked the shipped artifacts directly rather than going by the file listing of a local install. I replicated the selection logic in `getLatestCoreToolsRelease()` against the live feed (`cli-feed-v4.json`, keys reversed, stable 4.x only, `coreTools` entry matching the OS with `size: "full"`), then parsed each archive's ZIP central directory over HTTP range requests, so no full download was needed.

The release that selection resolves to today is 4.131.0:

* `Azure.Functions.Cli.linux-x64.4.13.0.zip`, 8703 central directory entries: `func` present at the root, no `gozip` at the root, `gozip` present at `in-proc6/gozip` and `in-proc8/gozip`.
* `Azure.Functions.Cli.osx-x64.4.13.0.zip`, 8698 central directory entries: identical in this respect.

So `path.join(dest, "gozip")` resolves to a path that no longer exists, while `func` is still where the code expects it.

Walking the stable 4.x Linux releases in the feed pins when this changed. The `in-proc6` and `in-proc8` folders appeared first and their `gozip` copies coexisted with the root copy for a long stretch; the root copy is what went away:

* 4.0.0 through 4.101.0: `gozip` only at the root, no `in-proc6` or `in-proc8` folder at all
* 4.102.0 through 4.126.0: `gozip` at the root **and** at `in-proc6/gozip` and `in-proc8/gozip`
* 4.127.0 onward, including the current 4.131.0: root copy gone, only `in-proc6/gozip` and `in-proc8/gozip`

That makes 4.127.0 the first release that breaks this code path.

## Why guarding the old path alone is not enough

The same central directory dump shows a second thing worth acting on: in every one of these archives, **every** entry has an external attribute field of `0`. Counted across all 8703 Linux entries and all 8698 macOS entries, zero entries carry a non zero value.

These packages therefore ship no Unix permission bits at all. That is exactly why the explicit chmod on `func` exists in the first place, and it means every other binary in the package also lands without the executable bit.

Wrapping the existing `path.join(dest, "gozip")` call in an `fs.existsSync` check would stop the crash, but it would leave the real binaries non executable. That trades a loud failure for a quiet one.

## Which binaries actually need the flag

The same 4.102.0 release that introduced `in-proc6/gozip` and `in-proc8/gozip` also introduced `in-proc6/func` and `in-proc8/func`. Those two are the in process host executables: each sits next to its own `func.dll`, `func.dll.config`, `createdump` and `libhostfxr.so`, which is the same self contained apphost layout as the package root where `func` is already chmodded today. Core Tools launches one of them for .NET in process function apps.

The authoritative reference for which files need the flag is Microsoft's own installer for the `azure-functions-core-tools` npm package, which unpacks the identical archive. Its `lib/install.js` at 4.13.0 does exactly this:

```js
if (platform === 'linux' || platform === 'darwin') {
    fs.chmodSync(`${installPath}/func`, 0o755);

    // inproc is not packaged in the linux-arm64 builds, so skip setting permissions for that platform
    if (!(platform === 'linux' && arch === 'arm64')) {
        fs.chmodSync(`${installPath}/in-proc8/func`, 0o755);
        fs.chmodSync(`${installPath}/in-proc6/func`, 0o755);
    }
}
```

So the set this file has to cover is `func` plus the two in process hosts, and the `gozip` copies it already tries to handle. Note also that upstream skips the in process folders entirely on builds that do not ship them, which is the behaviour the existence check below gives us for free.

## What the fix does

Enumerate the binaries that need the executable flag and chmod the ones that are actually on disk:

```ts
const EXECUTABLE_BINARIES = ["func", "gozip", "in-proc6/func", "in-proc6/gozip", "in-proc8/func", "in-proc8/gozip"];

// Fix permissions on MacOS/Linux
if (os.platform() === "linux" || os.platform() === "darwin") {
  for (const binary of EXECUTABLE_BINARIES) {
    const binaryPath = path.join(dest, binary);
    if (fs.existsSync(binaryPath)) {
      fs.chmodSync(binaryPath, 0o755);
    }
  }
}
```

Notes on the shape of this:

* The root `gozip` entry stays in the list on purpose. Releases up to 4.126.0 still ship it there, and this code path is reached for any 4.x version the feed resolves to, so dropping it would regress those.
* `fs.existsSync` guarding is the pattern already used elsewhere in this file, in `getDownloadedCoreToolsVersion()` and at the top of `downloadCoreTools()`. Here it also covers packages built without the in process folders, without needing to know the architecture.
* `EXECUTABLE_BINARIES` sits with the other module constants, and uses forward slashes the same way `CORE_TOOLS_FOLDER` does, since both are fed through `path.join`.

## Testing

The download path tests in `src/core/func-core-tools.spec.ts` were inert: the `adm-zip` mock was commented out, and both tests that exercise `downloadCoreTools()` were `it.skip`. The only mention of `gozip` in the file was inside the dead comment block.

I replaced that commented block with a working `adm-zip` mock. It writes a caller supplied set of files into memfs at the extraction destination using `vol.fromJSON`, which leaves them at memfs's default mode of `0o666`, matching the real archives that carry no permission bits. That default is what makes the mode assertions meaningful rather than vacuous.

Both previously skipped tests now work and are un skipped and passing. Three new tests cover the layouts:

* `should make the binaries executable when gozip sits at the package root` (the pre 4.127.0 layout)
* `should make the binaries executable when gozip sits in the in process host folders` (the current layout, asserting `func`, both `in-proc*/func` and both `in-proc*/gozip`)
* `should skip the in process host binaries when the package does not ship them` (a package with only a root `func`, which must still complete)

Result on the changed file:

```
 ✓ funcCoreTools > getCoreToolsBinary > should download core tools and return downloaded binary
 ✓ funcCoreTools > downloadCoreTools > should throw an error if the download is corrupted
 ✓ funcCoreTools > downloadCoreTools > should make the binaries executable when gozip sits at the package root
 ✓ funcCoreTools > downloadCoreTools > should make the binaries executable when gozip sits in the in process host folders
 ✓ funcCoreTools > downloadCoreTools > should skip the in process host binaries when the package does not ship them
```

### The new tests fail without the source change

Reverting only `src/core/func-core-tools.ts` and keeping the tests reproduces the reported crash:

```
FAIL  src/core/func-core-tools.spec.ts > funcCoreTools > downloadCoreTools > should make the binaries executable when gozip sits in the in process host folders
Error: ENOENT: no such file or directory, open '/home/user/.swa/core-tools/v4/gozip'
 ❯ Volume.chmodSync node_modules/memfs/src/volume.ts:1828:10
 ❯ Module.downloadCoreTools src/core/func-core-tools.ts:257:8
    255|   if (os.platform() === "linux" || os.platform() === "darwin") {
    256|     fs.chmodSync(path.join(dest, "func"), 0o755);
    257|     fs.chmodSync(path.join(dest, "gozip"), 0o755);
       |        ^
```

Dropping only `in-proc6/func` and `in-proc8/func` from `EXECUTABLE_BINARIES`, keeping everything else, fails the same test on the mode assertion with `expected 438 to be 493`, that is `0o666` against `0o755`. The in process host coverage is therefore load bearing on its own.

### Full suite

`npm test` goes from 488 passed / 14 skipped to 493 passed / 12 skipped, with no change in what fails.

Two suites already fail on unmodified `main` in my environment, both from running Node 25 against a repo that targets Node 18 (`.nvmrc`), and both are untouched by this change:

* `src/cli/index.spec.ts` fails to load, via `jsonwebtoken` to `jwa` to `buffer-equal-constant-time`, which reads `SlowBuffer.prototype`. `SlowBuffer` was removed in recent Node.
* `funcCoreTools > getCoreToolsBinary > should return the system binary if it's compatible` expects `isCoreToolsVersionCompatible(4, <node major>)` to be true, but the table in this file caps v4 at Node 22, so Node 25 returns false.

I confirmed both by stashing the change and rerunning: identical failure counts before and after. CI covers Node 18, 20 and 22 on Linux, macOS and Windows, which is where this should get its real check.

`npx tsc --noEmit` is clean and `npx prettier --check` passes on both changed files.


Om-singhaI:
Could someone take a look at this when you get a chance? I'm happy to rework it if the approach isn't right.

Om-singhaI:
Checking in on this one. It's a small change to which Core Tools binaries get `chmod` treatment, and both checks pass. Happy to rebase or split it up if that makes review easier.

